PR #79, opened 2022-04-05
Improve ea_current_to_profile output parser

atoomic:
When using some unsupported package the
ea_current_to_profile script is going to
add noise to the STDOUT... we need to improve
the output parsing logic to be able to get
a json file.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1393,15 +1393,7 @@ sub backup_ea4_profile () {
         return;
     }
 
-    my @cmd     = qw{ /usr/local/bin/ea_current_to_profile --target-os=AlmaLinux_8};
-    my $cmd_str = join( ' ', @cmd );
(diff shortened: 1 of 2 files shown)


elevate-cpanel
```diff
@@ -1416,6 +1408,35 @@ sub backup_ea4_profile () {
...
+    die qq[Unable to backup EA4 profile. Failure from $cmd_str] if $?;
+
+    my @lines = split( "\n", $output );
+
+    my $json;
+    if ( scalar @lines == 1 ) {
+        $json = $lines[0];
+    }
+    else {
+        foreach my $l ( reverse @lines ) {
```
toddr, line 1425: What are we doing with the rest of the output? throwing it out?

toddr: meaning... is it redundant?

Issue #25, opened 2022-02-25
Add a check/warning to confirm we are using the latest cpanel-elevate


atoomic:
The goal is to detect deprecated scripts and display a warning to the user requesting him to update it from the last stable state.
This should not be a blocker but probably a warning displayed just before giving the user a few seconds to abort the process

toddr:
validate the file is identical to the raw file from github. This should only check when you run --start (this is a blocker only check)

PR #81, opened 2022-04-06
Check for whether the running script is updated

sloanebernstein:
Allow the elevate-cpanel script to notify the administrator if a newer
version of the script is available. Also allow the admin to skip this
blocker if desired.

Implements #25.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -49,6 +49,7 @@ Helps to upgrade CentOS 7 cPanel servers to AlmaLinux 8.
        --status                       Check the current elevation status
        --clean                        Cleanup scripts and files created by elevate-cpanel
 
+       --skip-elevate-version-check   Skip the check for whether this script is up to date.
        --skip-cpanel-version-check    Skip the check for whether cPanel is up to date.
                                       This option is intended only for testing!
 
@@ -137,6 +138,7 @@ use Log::Log4perl qw(:easy);
 
 use Config;
 use Carp           ();
+use Digest::SHA    ();
 use File::Basename ();
 use File::Copy     ();
 use File::Find     ();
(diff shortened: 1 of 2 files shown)


elevate-cpanel
```diff
@@ -2756,6 +2759,64 @@ sub _blocker_ea4_profile( $self ) {
     EOS
 }
 
+sub _blocker_script_updated ($self) {
+
+    my ( $should_block, $blocker_text ) = ( undef, undef );
+
+    if ( !$self->getopt('skip-elevate-version-check') ) {
```
toddr, line 2766: can't you just return here?

elevate-cpanel
```diff
@@ -2756,6 +2759,64 @@ sub _blocker_ea4_profile( $self ) {
...
+
+        my ( $latest_checksum, $self_checksum ) = ( _latest_checksum(), _self_checksum() );
+        if ( !defined $latest_checksum ) {
+            $should_block = 1;
+            $blocker_text = "The latest version of elevate-cpanel could not be fetched for comparison.";
+        }
+        elsif ( !defined $self_checksum ) {
+            $should_block = 1;
+            $blocker_text = "This script is not installed at the expected location /scripts/elevate-cpanel.";
```
toddr, line 2775: Should be unreachable code as we do this check already earlier. Fine to leave.

elevate-cpanel
sloanebernstein: This is the legacy of previous revisions of this logic.

elevate-cpanel
```diff
@@ -2756,6 +2759,64 @@ sub _blocker_ea4_profile( $self ) {
...
+        }
+        else {
+            $should_block = $latest_checksum eq $self_checksum ? 0 : 1;
+            $blocker_text = <<~EOS if $should_block;
+            This script does not appear to be the newest available release.
+            Please download the latest version from the following URL and try again:
+
+            https://raw.githubusercontent.com/cpanel/elevate/release/elevate-cpanel
```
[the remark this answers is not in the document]
sloanebernstein: A self-update is the logical conclusion of needing to fetch the latest release at all. We should handle that in another issue/PR.

Issue #83, opened 2022-04-07
Send all output through log4perl?


sloanebernstein:
Currently, most logging to file is done by piping the output of a command through `tee(1)`, which is only invoked through the systemd service. This means that some output is not captured in the log file, such as the entirety of stage 1. It also makes logging of some events (_e.g._, use of `--continue` in #71) awkward and ad-hoc.

Since much of the internal diagnostic output is being routed through `Log::Log4perl`, it would be ideal to also route external program output through it and use a native appender to handle logging to file.

Issues:
- `ssystem` needs to call something that lets it capture output (_i.e._, not `system`).
- There doesn't seem to be an easy way to colorize output to a file via log4perl: `ScreenColoredLevels` won't write to a file, and `File` can't do colors.

toddr:
log4perl can provide different output per stream so we've got some stuff to work with there.

sloanebernstein:
For now, logging consistency is more important than fancy colored text, so this feature is likely being sacrificed.

PR #84, opened 2022-04-07
Append to log on --continue

sloanebernstein:
Echo the `Restarting install process` message to the log file, to indicate that an attempt to resume from an error has been made.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -231,15 +231,13 @@ sub _OPTIONS {
     return qw( service start clean continue manual-reboots status log check skip-cpanel-version-check skip-elevate-version-check );
 }
 
-my $logger;
-
 exit( __PACKAGE__->new(@ARGV)->run() // 0 ) if !caller;
 
 sub run ($self) {
 
     local $| = 1;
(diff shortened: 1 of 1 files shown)


elevate-cpanel
```diff
@@ -230,15 +230,13 @@ sub _OPTIONS {
...
-my $logger;
-
 exit( __PACKAGE__->new(@ARGV)->run() // 0 ) if !caller;
 
 sub run ($self) {
 
     local $| = 1;
 
-    $logger = _init_logger();
+    _init_logger();
```
toddr, line 239: does this release the logger?

toddr: Maybe change the return of _init_logger to be empty if you think it has no value?

sloanebernstein: Technically, `Log::Log4perl->init()` has a return value hard-coded to `1` via `Log::Log4perl::Config::_init`. I figure that this can be addressed if/when we change everything over in #83.

toddr:
Let's revisit this after completing #83

PR #86, opened 2022-04-11
Handle log output with log4perl

### LINKED ISSUE #71: add to log when running with --continue
specify in log that a continue has been triggered when it happens?

sloanebernstein:
Originally, logging to a file was done by piping screen output to
`tee(1)`. This prevented some output from being logged, as only the
systemd service did this. Instead, use a native log4perl file appender
to log to the file.

This eliminates colored output, since there is no direct way to save
color data to the file anymore. Restoring this feature will be explored
in a future issue.

Resolves #71 and #83.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -231,15 +231,13 @@ sub _OPTIONS {
     return qw( service start clean continue manual-reboots status log check skip-cpanel-version-check skip-elevate-version-check );
 }
 
-my $logger;
-
 exit( __PACKAGE__->new(@ARGV)->run() // 0 ) if !caller;
 
 sub run ($self) {
 
     local $| = 1;
 
-    $logger = _init_logger();
+    _init_logger();
 
     if ( $self->getopt('start') ) {
         die qq[Unsupported option with --start\n] if $self->getopt('continue') || $self->getopt('service');
@@ -2463,7 +2461,6 @@ sub has_blocker ( $self, $id, $msg ) {
     *** Elevation Blocker detected: ***
     $msg
     EOS
-    WARN('.');
 
     $self->{blockers} //= [];
     push $self->{blockers}->@*, $blocker;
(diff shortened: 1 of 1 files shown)


elevate-cpanel
```diff
@@ -3195,7 +3192,7 @@ sub install_cpanel_elevate_service {
         Type=simple
         # want to run it once per boot time
         RemainAfterExit=yes
-        ExecStart=/usr/bin/bash -c 'exec /usr/local/cpanel/scripts/elevate-cpanel --service 2>&1 | /usr/bin/tee -a $log_file'
```
toddr, line 3198: WTH? why was this bash??

elevate-cpanel
```diff
@@ -3296,10 +3293,29 @@ sub read_redhat_release() {
...
+
+    # emulate shell behavior of system()
+    if ( scalar @args == 1 && $args[0] =~ m/[\$&*(){}\[\]'";\\|?<>~`\n]/ ) {
+        unshift @args, qw(/usr/bin/bash -c);
+    }
+
+    my $program = shift @args;
+    my ( $callback_out, $callback_err ) = map {
+        my $label = $_;
+        Cpanel::IOCallbackWriteLine->new( sub ($line) { chomp $line; INFO("{$label} $line"); } )
```
toddr, line 3306: INFO autochomps.

sloanebernstein:
If logging to the systemd journal is not desirable, the `Screen` appender will need to be disabled.

elevate-cpanel
```diff
@@ -3296,10 +3293,29 @@ sub read_redhat_release() {
...
+        program      => $program,
+        args         => [@args],
+        stdout       => $callback_out,
+        stderr       => $callback_err,
+        timeout      => 0,
+        read_timeout => 0,
+    );
+    INFO();    # Buffer so they can more easily read the output.
+
+    return $? = $sr->CHILD_ERROR;    ## no critic qw(Variables::RequireLocalizedPunctuationVars) -- emulate return behavior of system()
```
toddr, line 3318: I'm not wild about us clobbering $? why would we want to set that?

sloanebernstein: The use of `system` already clobbered `$?`, so this just maintains "feature" parity. It doesn't need to stay.

PR #96, opened 2022-04-21
Restore colorized logging

### LINKED ISSUE #87: Restore colorized screen output
#86 removed colorized screen output as a side effect of converting logging to a file using native log4perl mechanisms. Colorized output is useful for users who may not fully understand how to read the existing output, so it is desirable to restore this feature. Doing this for direct invocation is easy enough, so output during `--check` and stage 1 of `--start` is trivial to restore, but all other output is not.

Instead of directly tailing the log file, one way of solving this might be to have the script invoked as a service use a socket or named pipe to relay the logging information to the script invoked with `--log` or otherwise listening to the service.

sloanebernstein:
Fixes #87.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -137,16 +137,17 @@ use parent qw( Cpanel::HelpfulScript );
 use Log::Log4perl qw(:easy);
 
 use Config;
-use Carp           ();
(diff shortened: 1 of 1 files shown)


atoomic (approved):
sweet

PR #103, opened 2022-06-17
Implement a --check-json which outputs the list of blockers as a JSON array

sloanebernstein:

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -229,7 +229,7 @@ use constant VETTED_YUM_REPO => qw{
 use constant REBOOT_NEEDED => 4242;    # just one unique id
 
 sub _OPTIONS {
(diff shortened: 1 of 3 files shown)


elevate-cpanel
```diff
@@ -557,17 +561,26 @@ sub do_check($self) {
     # cleanup the stage file
     unlink ELEVATE_STAGE_FILE;
 
-    if ($has_blockers) {
-        WARN( <<~'EOS' );
-        Please fix the detected issues before performing the elevation process.
-        EOS
+    if ($self->getopt('check')) {
+        if ($has_blockers) {
```
atoomic, line 565: Why adding a new `check-json` option and not using the current check to populate our generic json file with the blockers?
should not all we need is to add here something like?
```
update_stage_file( { 'blockers' => $self->{'blockers'} } );
```

Issue #102, opened 2022-06-17
/elevate-cpanel --check should abort earlier when run on a C8 like server


atoomic:
When detecting a major version >=8 we should abort earlier and make the error fatal, not a warning

```
╰─> cat /etc/redhat-release
AlmaLinux release 8.6 (Sky Tiger)
```

```
╰─> /scripts/elevate-cpanel --check
info [elevate-cpanel] Successfully verified signature for cpanel (key types: release, development).
* 17-17:14:05 (2465) [WARN] *** Elevation Blocker detected: ***
This installation of cPanel (11.105.9999.77) does not appear to be up to date. Please upgrade cPanel to a most recent version.

* 17-17:14:05 (2465) [WARN] *** Elevation Blocker detected: ***
This script is only designed to upgrade CentOS 7 to AlmaLinux 8

* 17-17:14:05 (2465) [WARN] *** Elevation Blocker detected: ***
You need to run CentOS 7.9 and later to upgrade AlmaLinux 8. You are currently using AlmaLinux v8.6.0

* 17-17:14:07 (2971) [ERROR] 509 package(s) installed from unsupported YUM repo 'baseos' from /etc/yum.repos.d/almalinux.repo
* 17-17:14:07 (2971) [ERROR] 331 package(s) installed from unsupported YUM repo 'appstream' from /etc/yum.repos.d/almalinux.repo
* 17-17:14:07 (2981) [WARN] Unsupported YUM repo enabled 'extras' without packages installed from /etc/yum.repos.d/almalinux.repo
* 17-17:14:07 (2981) [WARN] Unsupported YUM repo enabled 'epel-modular' without packages installed from /etc/yum.repos.d/epel-modular.repo
* 17-17:14:07 (2465) [WARN] *** Elevation Blocker detected: ***
One or more enabled YUM repo are currently unsupported.
You should disable these repositories and remove packages installed from them
before continuing the update.

Consider reporting this limitation to https://github.com/cpanel/elevate/issues


* 17-17:14:14 (3042) [ERROR] Could not read directory '/var/lib/yum': No such file or directory
* 17-17:14:14 (2465) [WARN] *** Elevation Blocker detected: ***
yum is not stable

* 17-17:14:15 (2879) [INFO] Checking if your system is up to date:
* 17-17:14:15 (3300) [INFO] Running: /usr/bin/yum clean all
* 17-17:14:15 (3301) [INFO]
* 17-17:14:15 (3311) [INFO] 113 files removed
[...]

PR #109, opened 2022-07-15
Don't allow starting or checking if not on CentOS 7

sloanebernstein:
Primary purpose of these is to implement #102, but the other commits are cleanups / QoL improvements.


toddr:
How will this behave when security advisor / analytics calls this and it's not on CentOS 7?

sloanebernstein:
Security Advisor guards running the command by way of `Cpanel::OS::can_be_elevated`. Analytics does not run the program; it reports data files if they exist.

elevate-cpanel
```diff
@@ -589,6 +591,20 @@ sub save_blocker_file ( $path, $stash ) {
     return 1;
 }
 
+sub bail_out_on_inappropriate_distro () {
```
toddr, line 594: Please use Cpanel::OS here.

PR #113, opened 2022-08-04
Re-work the BLS blocker

sloanebernstein:
QA reported that #112 was unable to handle double-quoted values, which
it should, because `/etc/default/grub` and similar files are able to be
any valid shell variable assignments, including non-constant expressions.

Since there is no better parser of shell scripts than the shell, use
that to extract the value from the file. Drop UID to `nobody` and run
`bash` in its restricted mode to try to prevent unintended side effects.

Since direct I/O isn't used anymore, switch the tests of the blocker
itself to use Test::MockFile, and write a test for the shell-based
parsing mechanism. This requires rewriting a few of the previous tests.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -149,11 +149,14 @@ use File::Copy      ();
 use File::Find      ();
 use File::Path      ();
 use File::Slurper   ();
+use File::Spec      ();
 use File::Temp      ();
 use Hash::Merge     ();
 use LWP::Simple     ();
 use Term::ANSIColor ();
 
+use Cpanel::AccessIds::SetUids ();
+use Cpanel::Binaries ();
 use Cpanel::Config::Httpd                       ();
 use Cpanel::Config::LoadCpConf                  ();
 use Cpanel::HTTP::Client                        ();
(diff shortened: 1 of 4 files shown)


elevate-cpanel
```diff
@@ -2896,13 +2898,7 @@ sub _latest_checksum () {
...
     if ( !$self->getopt('skip-disable-blscfg') ) {
-        my $etc_default_grub = Cpanel::Transaction::File::LoadConfigReader->new(
-            path      => DEFAULT_GRUB_FILE,
-            delimiter => '=',
-            comment   => '^\s*#',
-        );
-
-        my $grub_enable_blscfg = $etc_default_grub->get_entry('GRUB_ENABLE_BLSCFG');
+        my $grub_enable_blscfg = _parse_shell_variable( DEFAULT_GRUB_FILE, 'GRUB_ENABLE_BLSCFG' );
         return $self->has_blocker( 106, <<~EOS ) if !$grub_enable_blscfg || $grub_enable_blscfg ne 'false';
```
toddr, line 2902: eq true?

sloanebernstein: From an AL8 instance:
```
[root@10-2-64-119 ~]# grep BLSCFG /etc/grub.d/10_linux
  if [ "x${GRUB_ENABLE_BLSCFG}" = "xtrue" ]; then
if [ "x${GRUB_ENABLE_BLSCFG}" = "xtrue" ]; then
  if [ "x${GRUB_ENABLE_BLSCFG}" != "xtrue" ]; then
    if [ "x${GRUB_ENABLE_BLSCFG}" != "xtrue" ]; then
```

PR #118, opened 2022-08-31
Address BLS by detecting and updating provider workarounds for GRUB2 prefix

sloanebernstein:
At least one server provider of note boots their default Linux images using a version of GRUB2 which looks for its configuration, boot entries, and modules at `/boot/grub`, which is the Debian convention; however, the RHEL convention is to install GRUB2 things into `/boot/grub2`. In images for older distros (such as CentOS 7), they had been working around this by creating a symlink at `/boot/grub/grub.cfg` which points to `/boot/grub2/grub.cfg`, which allows booting when the boot entries are in native GRUB2 configuration. However, in newer distros, they have changed this workaround to symlink the entirety of `/boot/grub` to point to `/boot/grub2`. The reason for this is that BLS is implemented through the use of a GRUB2 module, and the symlink of only the configuration file neglects to give their instance of GRUB2 access to load any of the modules installed on the operating system's filesystem, while the symlink of the entire directory does grant this access.

Attempt to detect this workaround based on the types and structure of these files, directories, and symlinks. Block if we cannot clearly identify whether or not one of these workarounds is in place. Update the workaround during stage 2 if the older variant is found and, as a courtesy, try to copy non-conflicting files, etc. from the old `/boot/grub` into `/boot/grub2`.

Additionally, fix what is probably a mistake in the names of the entries in the stash file created by `run_once()`.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -141,18 +141,22 @@ use parent qw( Cpanel::HelpfulScript );
 use Log::Log4perl qw(:easy);
 
 use Config;
-use Carp            ();
-use Digest::SHA     ();
-use File::Basename  ();
-use File::Copy      ();
-use File::Find      ();
-use File::Path      ();
-use File::Slurper   ();
-use File::Temp      ();
-use Hash::Merge     ();
-use LWP::Simple     ();
-use Term::ANSIColor ();
+use Carp                  ();
+use Digest::SHA           ();
+use File::Basename        ();
+use File::Copy            ();
+use File::Copy::Recursive ();
+use File::Find            ();
+use File::Path            ();
+use File::Slurper         ();
+use File::Spec            ();
+use File::Temp            ();
+use Hash::Merge           ();
+use LWP::Simple           ();
+use Term::ANSIColor       ();
 
+use Cpanel::AccessIds::SetUids  ();
+use Cpanel::Binaries            ();
 use Cpanel::Config::Httpd       ();
 use Cpanel::Config::LoadCpConf  ();
 use Cpanel::HTTP::Client        ();
@@ -163,6 +167,7 @@ use Cpanel::Pkgr                ();
 use Cpanel::RestartSrv::Systemd ();
 use Cpanel::SafeRun::Simple     ();
 use Cpanel::SafeRun::Errors     ();
+use Cpanel::SafeRun::Object     ();
 use Cpanel::Update::Tiers       ();
 use Cpanel::Version::Tiny       ();
 use Cpanel::Version::Compare    ();
@@ -194,6 +199,8 @@ use constant IMUNIFY_LICENSE_BACKUP => ELEVATE_BACKUP_DIR . '/imunify-backup-lic
 use constant SBIN_IP => q[/sbin/ip];
 
 use constant YUM_REPOS_D => q[/etc/yum.repos.d];
+
+use constant DEFAULT_GRUB_FILE => '/etc/default/grub';
 
 use constant DISABLE_MYSQL_YUM_REPOS => qw{
   Mysql57.repo
@@ -232,6 +239,13 @@ use constant VETTED_YUM_REPO => qw{
 }, DISABLE_MYSQL_YUM_REPOS;
 
 use constant REBOOT_NEEDED => 4242;    # just one unique id
+
+use constant {
+    GRUB2_WORKAROUND_NONE      => 0,
+    GRUB2_WORKAROUND_OLD       => 1,
+    GRUB2_WORKAROUND_NEW       => 2,
+    GRUB2_WORKAROUND_UNCERTAIN => -1,
+};
 
 sub _OPTIONS {
     return qw( service start clean continue manual-reboots status log check:s skip-cpanel-version-check skip-elevate-version-check );
@@ -694,6 +708,9 @@ sub run_stage_2 ($self) {
     setup_motd();
 
     run_once('remove_kernelcare_if_needed');
+
+    run_once('update_grub2_workaround_if_needed');    # required part
+    run_once('merge_grub_directories_if_needed');     # best-effort part
 
     return REBOOT_NEEDED;
 }
@@ -2472,7 +2489,7 @@ sub blockers_check ( $self, $all = 0 ) {
             ERROR( $error->{msg} );
             return $error->{id} // 401;
         }
-        WARN("Unkown error while checking blockers: $error");
+        WARN("Unknown error while checking blockers: $error");
         return 127;    # unknown error
     }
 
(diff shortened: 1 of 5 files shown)


sloanebernstein:
Note: This PR is against the temporary `bls` branch. The new code builds upon some of the changes made there, such as inverting the logic of the old BLS blocker, such that if `GRUB_ENABLE_BLSCFG` is present, it must be set to `true`, or the script will block. As such, even when this is merged, the `bls` branch will need to be merged into `main`.

elevate-cpanel
```diff
@@ -2897,61 +2907,224 @@ sub _latest_checksum () {
...
-        This system is currently configured so that the entries for the boot loader
-        will be stored in BLS format upon upgrade. This is ordinarily fine, but some
-        providers boot into their own copy of GRUB2, and theirs may not understand the
-        BLS format. If this is the case, the system will not be able to boot through
-        normal means in the middle of the upgrade process.
+    my $grub_enable_blscfg = _parse_shell_variable( DEFAULT_GRUB_FILE, 'GRUB_ENABLE_BLSCFG' );
 
-        The safe option is to add or change the following line in /etc/default/grub so
-        that it reads:
+    return $self->has_blocker( 106, <<~EOS ) if defined $grub_enable_blscfg && $grub_enable_blscfg ne 'true';
```
toddr, line 2912: Who are we trying to protect ourselves from?

elevate-cpanel
```diff
@@ -2897,61 +2907,224 @@ sub _latest_checksum () {
...
 }
 
+sub GRUB2_PREFIX_DEBIAN { return '/boot/grub' }
+sub GRUB2_PREFIX_RHEL   { return '/boot/grub2' }
+
+sub _blocker_grub2_workaround ($self) {
+    my $state = _grub2_workaround_state();
+    if ( $state == GRUB2_WORKAROUND_OLD ) {
+        my ( $deb, $rhel ) = ( GRUB2_PREFIX_DEBIAN, GRUB2_PREFIX_RHEL );
+        WARN( <<~EOS );    # FIXME: should this be WARN or INFO?
```
toddr, line 2970: WARN works. no need to fixme.

elevate-cpanel
```diff
@@ -2897,61 +2907,224 @@ sub _latest_checksum () {
...
+                }
+
+                if ( !File::Copy::Recursive::rcopy( "$grub_bak/$entry", "$grub2_dir/$entry" ) ) {
+                    $failed_copy++;
+                    WARN("Copying \"$grub_bak/$entry\" into \"$grub2_dir\" failed.");
+                }
+            }
+            closedir $grub_bak_fh;
+        };
+        WARN("Unable to copy the contents of \"$grub_bak\" into \"$grub2_dir\": $@") if $@;
```
toddr, line 3122: Does `File::Copy::Recursive::rcopy` die?

elevate-cpanel
```diff
@@ -2897,61 +2907,224 @@ sub _latest_checksum () {
...
+            while ( my $entry = readdir $grub_bak_fh ) {
+
+                next if $entry eq '.';
+                next if $entry eq '..';
+
+                # grub.cfg in the backup dir is the old symlink, so ignore it separately to avoid warnings:
+                next if $entry eq "grub.cfg";
+
+                if ( exists $grub2_entries{$entry} ) {
+                    $skipped_copy++;
```
atoomic, line 3110: is it used somewhere?

toddr, line 3110: what do we do with `$skipped_copy`?

sloanebernstein: Oh, right.

I think I originally intended to add that info to the stage file for a later recap.

elevate-cpanel
toddr: doesn't this code just go away when merging to main?

elevate-cpanel
sloanebernstein: It isn't documented to do so, nor do the functions it mentions it calls. I can verify this if wanted, of course.

elevate-cpanel
toddr: needs to merge to main pls. bls was just for you to reference.

elevate-cpanel
```diff
@@ -2897,61 +2907,224 @@ sub _latest_checksum () {
...
+
+        If your GRUB2 configuration has not been customized, consider reporting
+        this limitation to https://github.com/cpanel/elevate/issues
+        EOS
+    }
+
+    return 0;
+}
+
+# TODO: This should probably return warnings/errors rather than directly printing to the log.
```
toddr, line 2999: Outstanding TODO?

elevate-cpanel
sloanebernstein: `opendir or die`

elevate-cpanel
sloanebernstein: It's more "something that could be nicer to work with" TODO than "thing I absolutely need to implement" TODO.

elevate-cpanel
```diff
@@ -2888,6 +2905,238 @@ sub _latest_checksum () {
...
+sub _blocker_blscfg ($self) {
+
+    my $grub_enable_blscfg = _parse_shell_variable( DEFAULT_GRUB_FILE, 'GRUB_ENABLE_BLSCFG' );
+
+    return $self->has_blocker( 106, <<~EOS ) if defined $grub_enable_blscfg && $grub_enable_blscfg ne 'true';
+    Disabling the BLS boot entry format prevents the resulting system from
+    adding kernel updates to any boot loader configuration, because the old
+    utility responsible for maintaining native GRUB2 boot loader entries was
+    removed and replaced with a wrapper around the new utility, which only
+    understands BLS format. This means that the old kernel will be used on
```
sloanebernstein, line 2917: BLS blocker message changed to no longer mention previous versions.

elevate-cpanel
```diff
@@ -2888,6 +2905,238 @@ sub _latest_checksum () {
...
+        add_final_notification( <<~EOS ) if ( $skipped_copy > 0 || $failed_copy > 0 );
+        After converting "$grub_dir" from a directory to a symlink to
+        "$grub2_dir", the upgrade process chose not to copy $skipped_copy
+        entries from the old directory due to name conflicts, and it failed to
+        copy $failed_copy entries due to system errors. The previous contents
+        of "$grub_dir" are now located at "$grub_bak". See $log_file for
+        further information.
+
+        If you did not add these files or otherwise customize the boot loader
+        configuration, you may ignore this message.
```
toddr, line 3132: I'm not sure about "you may ignore this message." but I'm not sure what alternate language to suggest.

sloanebernstein: My thinking here is that if not copying over a file breaks something non-critical with the bootloader, and the admin didn't take the action which caused the dependence on that file, there's no useful advice to give, thus it's better not to suggest anything to make it worse or otherwise confuse them. The final notification serves as notice to anyone who did perform weird customization that they should look to the log file to find out what files and issues occurred if something ancillary or cosmetic with the bootloader broke, rather than leaving them in the dark.

PR #127, opened 2022-09-29
Remove backup EA4 profile if run with --check

### LINKED ISSUE #126: elevate-cpanel --check is not cleaning up the EA4 profile it creates
The `elevate-cpanel` script presently generates a backup of the current EasyApache 4 profile during the evaluation of blockers. It is appropriate to retain this profile when evaluating prior to a real upgrade attempt, but it is not appropriate when only blockers are being evaluated via the `--check` flag. In combination with [the change to cPanel version 106 which runs the script in check mode as part of the Security Advisor](https://docs.cpanel.net/changelogs/106-change-log/#106-0-1), this is spamming the list of EA4 profiles with useless entries.

sloanebernstein:
Prevent unwanted retention of the created file when the script has no intention of performing the upgrade.

Fixes #126 and CPANEL-41603.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1558,17 +1558,20 @@ sub backup_ea4_profile ($self) {
         return;
     }
 
-    my $json = $self->_get_ea4_profile();
+    my $json_path = $self->_get_ea4_profile();
 
-    my $data = { profile => $json };
+    my $data = { profile => $json_path };
 
     # store dropped packages
-    my $profile = eval { Cpanel::JSON::LoadFile($json) } // {};
+    my $profile = eval { Cpanel::JSON::LoadFile($json_path) } // {};
     if ( ref $profile->{os_upgrade} && ref $profile->{os_upgrade}->{dropped_pkgs} ) {
         $data->{dropped_pkgs} = $profile->{os_upgrade}->{dropped_pkgs};
     }
 
     update_stage_file( { ea4 => $data } );
+
+    # CPANEL-41603: delete ea4 backup file if in --check mode
+    unlink $json_path unless $self->{_abort_on_first_blocker};
 
     return 1;
 }
--- a/t/ea4.t
+++ b/t/ea4.t
@@ -76,7 +76,7 @@ sub teardown : Test( teardown => 1 ) ($self) {
 }
 
 sub cpev {    # helper for test...
-    return bless {}, 'cpev';
+    return bless {@_}, 'cpev';
 }
 
 sub test_backup_and_restore_not_using_ea4 : Test(7) ($self) {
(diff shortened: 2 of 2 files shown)


elevate-cpanel
```diff
@@ -2820,7 +2820,9 @@ sub _blocker_ea4_profile ($self) {
 
     INFO("Checking EasyApache profile compatibility with AlmaLinux 8.");
 
-    return unless backup_ea4_profile();
+    my $backup = backup_ea4_profile();
+    return         unless $backup;
+    unlink $backup unless $self->{_abort_on_first_blocker};
```
atoomic, line 2825: this is correct but this logic should belong to `backup_ea4_profile` itself
and we should convert as a method and call it `$self->backup_ea4_profile`

elevate-cpanel
```diff
@@ -1456,7 +1456,7 @@ sub backup_ea4_profile () {
 
     update_stage_file( { ea4 => $data } );
 
-    return 1;
+    return $json;
```
atoomic, line 1459: note: the variable name `$json` is confusing, this is not part of these  changes
but could good to rename it as `ea_json_profile_file` or similar so we know this is a file

atoomic (changes requested):

elevate-cpanel
sloanebernstein: `backup_ea4_profile` is now a method.

elevate-cpanel
sloanebernstein: Changed to `$json_path`, if that's any better.

atoomic:
note: conflict due to multiple functions which were converted to methods to add support to Rocky

elevate-cpanel
```diff
@@ -1558,18 +1558,21 @@ sub backup_ea4_profile ($self) {
...
+    my $profile = eval { Cpanel::JSON::LoadFile($json_path) } // {};
     if ( ref $profile->{os_upgrade} && ref $profile->{os_upgrade}->{dropped_pkgs} ) {
         $data->{dropped_pkgs} = $profile->{os_upgrade}->{dropped_pkgs};
     }
 
     update_stage_file( { ea4 => $data } );
 
+    # CPANEL-41603: delete ea4 backup file if in --check mode
+    unlink $json_path unless $self->{_abort_on_first_blocker};
+
```
atoomic, line 1575: do we need to store the profile path in data when `_abort_on_first_blocker`?

t/ea4.t
```diff
@@ -322,6 +326,31 @@ Please remove these packages before continuing the update.
...
+sub test_blocker_file_behavior : Tests(8) ($self) {
+
+    my $ea_info_check = sub {
+        message_seen( 'INFO' => "Checking EasyApache profile compatibility with AlmaLinux 8." );
+    };
+
+    $self->{mock_cpev}->redefine( _get_ea4_profile => sub { $self->{mock_profile}->contents('{}'); return PROFILE_FILE } );
+
+    my $cpev = bless {}, 'cpev';
+    $cpev->{_abort_on_first_blocker} = 1;    # enforce a die
```
atoomic, line 338: note this can also be replaced by: `my $cpev = cpev(_abort_on_first_blocker => 1 )`

atoomic (approved):

t/ea4.t
atoomic: applied via e43b899ec861d1d000e3779bebf37075ccd4f82b

PR #132, opened 2022-10-06
Implement self-update mechanism

sloanebernstein:
Allow users to update to the latest version of the script using the `--update` flag.

Due to the need for an easy way to identify whether the existing version of the script on disk has been superseded by a newer one, an explicit version number is being introduced. Incrementing this is to be performed by a GitHub Action whenever the `release` tag is changed. Both the `elevate-cpanel` script and the `version` file will be changed to have the same value.


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,39 @@
...
+        submodules: true  # Fetch Hugo themes (true OR recursive)
+        fetch-depth: 0    # Fetch all history for .GitInfo and .Lastmod
+
+    - name: bump version number
+      run: |
+        set -x
+        VERSION=$(cat ./version)
+        let ++VERSION
+        echo -n $VERSION > ./version
+        sed -i -re "/^#<<V/,+1 s/VERSION => [0-9]*;/VERSION => ${VERSION};/" ./elevate-cpanel
```
toddr, line 28: This needs to go in a Makefile (make release) not in a workflow.

version
```diff
@@ -0,0 +1 @@
+0
```
toddr, line 1: Please start with 1

elevate-cpanel
```diff
@@ -173,6 +173,12 @@ use Cpanel::Version::Tiny       ();
 use Cpanel::Version::Compare    ();
 use Cpanel::Yum::Vars           ();
 
+#<<V *** DO NOT EDIT THESE LINES MANUALLY ***
+use constant VERSION => 0;
```
toddr, line 177: Why a constant for $VERSION?

version
sloanebernstein: It starts with 0 in the commit so that when the commit set is merged to main, the workflow will increase it to 1.

elevate-cpanel
```diff
@@ -1952,21 +1991,21 @@ sub _fetch_imunify_installer ($product) {
         return;
     };
 
-    return _fetch_a_script( $installer_url, 'imunify_installer' );
+    return _fetch_a_script( $installer_url, 'imunify_installer', '.sh' );
 }
 
-sub _fetch_a_script ( $url, $template ) {
+sub _fetch_a_script ( $url, $template, $suffix = '' ) {
```
toddr, line 1997: why not default suffix to .sh since that's what every existing caller used?

elevate-cpanel
sloanebernstein: Does this not resolve to the same thing as `sub VERSION { ... }`? Or would you prefer the `our $VERSION = ...` convention?

elevate-cpanel
sloanebernstein: I was going off of URLs, and this didn't seem to warrant a `.sh`, even though it got one:

```perl
      my $installer_script = _fetch_a_script( 'https://kernelcare.com/installer', 'kernelcare_installer' );
```

However, on doing an HTTP request, I see that URL redirects to something that does end in `.sh`:

```
# curl -LI https://kernelcare.com/installer
HTTP/1.1 301 Moved Permanently
Date: Thu, 06 Oct 2022 23:42:43 GMT
Content-Type: text/html
Connection: keep-alive
Location: https://repo.cloudlinux.com/kernelcare/kernelcare_install.sh
```

.github/workflows/bump-version.yml
toddr: I'm not ok with a version bump except when we do a release. Let's make this a Makefile

sloanebernstein: Version bump is now accomplished via Makefile:
```
[root@sandbox elevate]# make -n release
echo 1 > version
sed -i -re "/^#<<V/,+1 s/VERSION => [0-9]*;/VERSION => 1;/" elevate-cpanel
/usr/local/cpanel/3rdparty/bin/git commit -m "Release version 1" -- version elevate-cpanel
/usr/local/cpanel/3rdparty/bin/git tag -f release
/usr/local/cpanel/3rdparty/bin/git push origin
/usr/local/cpanel/3rdparty/bin/git push --force origin tag release
```

Makefile
```diff
@@ -17,3 +20,12 @@ tags:
...
 clean:
 	rm -f tags
+
+release: version := $(shell dc -f version -e '1 + p')
+release:
+	echo ${version} > version
+	sed -i -re "/^#<<V/,+1 s/VERSION => [0-9]*;/VERSION => ${version};/" elevate-cpanel
+	$(GIT) commit -m "Release version ${version}" -- version elevate-cpanel
+	$(GIT) tag -f $(RELEASE_TAG)
+	$(GIT) push origin
```
toddr, line 30: won't this fail?

sloanebernstein:
I've already found a bug in my implementation, but that's being addressed as part of another PR.

PR #133, opened 2022-10-07
read_stage file can read a specific data

atoomic:
read_stage return the whole stash but most of the
time we need to read one piece of it.

This change teaches 'read_stage' to take an optional argument and return the values for that key.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -435,9 +435,7 @@ sub check_status ($self) {
         return;
     }
 
-    my $stage_info = read_stage_file();
-
-    my $status = $stage_info->{status} // 'unknown';
+    my $status = read_stage_file('status') // 'unknown';
 
     if ( $status eq 'success' ) {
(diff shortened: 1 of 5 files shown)


elevate-cpanel
```diff
@@ -3536,8 +3525,15 @@ sub install_cpanel_elevate_service {
 
 # TODO: We're going to need to store reboot time so we know if the last reboot has happened when we re-run the script.
 
-sub read_stage_file {
-    return eval { Cpanel::JSON::LoadFile(ELEVATE_STAGE_FILE) } // {};
+sub read_stage_file ( $k = undef ) {
+    my $stage_info = _read_stage_file() // {};
+
+    return $stage_info->{$k} if defined $k;
```
toddr, line 3531: it's not a big deal but I wonder if you should be stringifying it? 
`$stage_info->{"$k"}`

PR #141, opened 2022-10-11
Allow alternate update URL

sloanebernstein:
For testing purposes, allow overriding the base URL for version checking and updates without modifying the script. Additionally, fix a bug in the update process that prevents the script from properly detecting that it is up to date.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -191,7 +191,7 @@ use Cpanel::Yum::Vars           ();
 use constant VERSION => 1;
 #>>V *** DO NOT EDIT THESE LINES MANUALLY ***
 
-use constant ELEVATE_BASE_URL => 'https://raw.githubusercontent.com/cpanel/elevate/release/';
+use constant DEFAULT_ELEVATE_BASE_URL => 'https://raw.githubusercontent.com/cpanel/elevate/release/';
 
 use constant CHKSRVD_SUSPEND_FILE => q[/var/run/chkservd.suspend];
 use constant ELEVATE_BLOCKER_FILE => '/var/cpanel/elevate-blockers';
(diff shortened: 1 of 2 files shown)


elevate-cpanel
```diff
@@ -388,6 +388,10 @@ sub do_update ($self) {
     return 0;
 }
 
+sub base_url ($self) {
+    return $self->{'base_url'} ||= $ENV{'ELEVATE_BASE_URL'} || DEFAULT_ELEVATE_BASE_URL;
```
atoomic, line 392: not really sure how will behave the version compared with different BASE_URL... 
but ok

atoomic (approved):

elevate-cpanel
sloanebernstein: Use case: We decide to change something about the way updates are detected and handled. We need to QA-test this before publishing to `release`. QA (or someone on their behalf) can create a branch off of `main` and both bump the version on this as well as create their own testing tag in place of `release` using `make RELEASE_TAG=tag release`. Before this change, one would need to hand-edit the script to point to this different URL, but now it can be specified during the actual test as an environment variable. (I could have made it an undocumented flag, I suppose.)

This mirrors the way that I tested the creation of the update process itself, by creating branch `test-make-release` and tag `test-tag`. (I've kept those around for now, but I'll delete them once we make the next release.)

elevate-cpanel
```diff
@@ -3035,7 +3039,7 @@ sub is_script_out_of_date ($self) {
         $blocker_text = "The script could not fetch information about the latest version.";
     }
     else {
-        $should_block = $latest_version eq $self_version ? 0 : 1;
+        $should_block = $latest_version == $self_version ? 0 : 1;
```
toddr, line 3042: I'm concerned that we're not validating `$latest_version` and `$self_version` are integers before doing an == check.

sloanebernstein: `$self_version` is just the `VERSION` constant, so we can guarantee it's an integer.
`$latest_verson` eventually calls `_latest_version()` as a builder:

```perl
sub _latest_version ($self) {
    my $http = Cpanel::HTTP::Client->new->die_on_http_error();
    my $url  = $self->ELEVATE_BASE_URL;
    my $response = $http->get( $url . '/version' );
    return int $response->content();
}
```

This dies on errors, and it relies on the URL returning something that can be interpreted by `int()`. In principle, a rogue proxy could cause a client to see the empty string.

PR #157, opened 2022-11-30
Do not block when detecting kernel update required

### LINKED ISSUE #130: Tolerate out of date kernel for upgrade check.
--check seems to be throwing a blocker if the system is not up to date. Does it need to do this? We auto-upgrade anyway before we run elevate? If nothing else it shouldn't block on kernels being out of date given cPanel usually blocks this for automatic upgrades.

atoomic:
Fix #130

Adjust the '_system_update_check' to do not block
when only kernel updates are required from:
	yum check-update

If any other packages than a 'kernel*' is detected then this is still a blocker and we request the use to then run 'yum update' as before.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -3338,9 +3338,25 @@ sub _system_update_check() {
 
     INFO("Checking if your system is up to date: ");
     ssystem(qw{/usr/bin/yum clean all});
-    if ( ssystem(qw{/usr/bin/yum check-update}) != 0 ) {
+
+    my $out = ssystem_capture_output(qw{/usr/bin/yum check-update -q});
+
+    if ( $out->{status} != 0 ) {
+
+        # not a blocker: only a warning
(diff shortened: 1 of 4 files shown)


elevate-cpanel
```diff
@@ -3338,9 +3338,25 @@ sub _system_update_check() {
...
     INFO("Checking if your system is up to date: ");
     ssystem(qw{/usr/bin/yum clean all});
-    if ( ssystem(qw{/usr/bin/yum check-update}) != 0 ) {
+
+    my $out = ssystem_capture_output(qw{/usr/bin/yum check-update -q});
+
+    if ( $out->{status} != 0 ) {
+
+        # not a blocker: only a warning
         WARN("Your system is not up to date please run: /usr/bin/yum update");
```
toddr, line 3347: I don't understand why this isn't a blocker. doesn't leapp freak out if you're not up to date on 7 before you go to 8?

Issue #190, opened 2023-02-07
eth0 can be renamed during elevate process leading to reboot issues


atoomic:
We can see that elevate renamed the `eth0` device
```
/var/log/leapp/leapp-upgrade.log:2023-01-02 20:45:01.254 DEBUG    PID: 19104 leapp.workflow.FactsCollection.udevadm_info: E: DEVPATH=/devices/pci0000:00/0000:00:12.0/virtio3/net/eth0
/var/log/leapp/leapp-upgrade.log:2023-01-02 20:45:01.384 DEBUG    PID: 19104 leapp.workflow.FactsCollection.udevadm_info: E: INTERFACE=eth0
/var/log/leapp/leapp-upgrade.log:2023-01-02 20:45:01.399 DEBUG    PID: 19104 leapp.workflow.FactsCollection.udevadm_info: E: SYSTEMD_ALIAS=/sys/subsystem/net/devices/eth0
/var/log/leapp/leapp-upgrade.log:2023-01-02 20:45:22.58  INFO     PID: 19448 leapp.workflow.FactsCollection.persistentnetnamesdisable: Single eth0 network interface detected. Appending 'net.ifnames=0' to RHEL-8 kernel commandline
/var/log/leapp/leapp-upgrade.log:2023-01-02 22:53:08.842 WARNING  PID: 698 leapp.workflow.Applications.persistentnetnamesconfig: Detected interface rename eth0 -> ens18.
/var/log/leapp/leapp-upgrade.log:Jan 02 20:39:37 localhost kernel: virtio_net virtio3 ens18: renamed from eth0
```

We should at minimum raise a warning at the end of the elevation process when detecting interface renamed like above.
As this could lead to some issues while rebooting the server later.

```
No suitable device found for this connection (device ens18 not available because profile is not compatible with device
```

Internal Ticket: #94525318

toddr:
What could go wrong if the interfaces are NOT renamed?

toddr:
What we think is happening: 

When leapp detects a single eth0 it does:
```
2023-01-02 20:45:22.58  INFO     PID: 19448 leapp.workflow.FactsCollection.persistentnetnamesdisable: Single eth0 network interface detected. Appending 'net.ifnames=0' to RHEL-8 kernel commandline
```

which leads to this:

```
2023-01-02 22:54:30.39  INFO     PID: 3 leapp.workflow.Finalization: Executing actor kernelcmdlineconfig 
[...]

PR #193, opened 2023-02-08
Avoid renaming network interfaces

atoomic:
Refs: #190

Setting that environment variable to 1
while running the leapp upgrade would
avoid renaming interfaces which could lead
to a server being unavaiable.


toddr:
I would like to get clarification from @prilr before we make this change.

toddr:
Remove `/var/cpanel/globalcache/cpanel.cache` after reboot. Update /etc/wwwacct.conf and caches.

atoomic:
we should not need to set that ENV variable

PR #195, opened 2023-02-08
Check and fix grub for net.ifnames=0

atoomic:
https://github.com/cpanel/elevate/issues/190

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1061,6 +1061,8 @@ sub run_stage_5 ($self) {
     File::Path::remove_tree(ELEVATE_BACKUP_DIR) if -d ELEVATE_BACKUP_DIR;
     cleanup_motd();
 
+    $self->check_and_fix_grub();
+
     my $final_notifications = read_stage_file( 'final_notifications', [] );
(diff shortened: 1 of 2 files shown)


toddr:
We shouldn't be doing this unless the new kernel has this option on it.

toddr (changes requested):
I'd rather wait for upstream if they can fix this quickly.

elevate-cpanel
```diff
@@ -1087,6 +1089,27 @@ sub run_stage_5 ($self) {
...
+    # for an autofixer
+    # return unless -e q[/var/cpanel/version/elevate];
+
+    my $grub_conf;
+    {
+        local $/;
+        open( my $fh, '<', DEFAULT_GRUB_FILE ) or return;
+        $grub_conf = <$fh>;
+    }
+
```
atoomic, line 1103: maybe check if `eth0` is here or the output from the leapp log to confirm something like:
```
Detected interface rename eth0 -> ens18.
```

atoomic: even better check the output from `/proc/cmdline`

PR #200, opened 2023-02-24
Exempt EA4 OpenSSL devel packages from blocker

### LINKED ISSUE #119: The system detected the following issues which would prevent cPanel ELevate from upgrading the system to AlmaLinux 8
Cloud KVM CentOS v7.9.2009 STANDARD kvm

One or more EasyApache 4 package(s) are not compatible with AlmaLinux 8. Please remove these packages before continuing the update. - ea-openssl-devel - ea-php74-php-ioncube12 - ea-php81-php-ioncube12

sloanebernstein:
The EasyApache packages for the development files belonging to the OpenSSL builds used on CentOS 7 do not depend on the main packages being present. As such, it is possible for the `ea-openssl-devel` package to have become an orphan when `ea-openssl` was replaced with `ea-openssl11`.

Since the absence of these development packages do not cause problems for cPanel itself, allow the script to ignore the removal of these packages, at least for the purpose of the EA4 compatibility blocker.

Fixes #119.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -3285,7 +3285,8 @@ sub _blocker_ea4_profile ($self) {
     my @incompatible_packages;
     foreach my $pkg ( sort keys $dropped_pkgs->%* ) {
         my $type = $dropped_pkgs->{$pkg} // '';
-        next if $type eq 'exp';    # use of experimental packages is a non blocker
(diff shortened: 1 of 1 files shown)


toddr:
I'm against working around bugs. We should get EA to fix this, possibly with the use of obsoletes/conflicts in their spec files.

Issue #198, opened 2023-02-22
Need a manual update blocker 


jasonmpotter:
I work as a Senior Linux SA with a major cPanel partner and we are starting to see customers shooting themselves in the foot trying to upgrade their production systems without consulting support first. 

As a matter of policy, we do not offer OS upgrades due to the various headaches they can cause for seasoned technical staff let alone less sophisticated customers who trust cPanel and see the notices for ELevate as a solution they can trust. So they charge forward and inadvertently brick their systems. 

I would like to see a manual update blocker built-in to the ELevate project. Just a simple /etc/disableelevate or similar that will cause the script to abort early with the failure message like: 

ELevate has been disabled on this server. Please contact your server provider support for assistance with upgrading to Alma/Rocky Linux.

It's a low lift solution that we can't get out in the wild fast enough as it can save many hours of frustration on both sides of the technical support coin. Thanks for your time.

jasonmpotter:
Even better, have the failure message be read right out of the /etc/disableelevate blocker file. Then hosting providers of all brands can manage the message they want to display to customers.

jasonmpotter:
After more digging into this feature request, I was able to partially implement in on our end, but it relies on making the blocker files immutable. This will suffice, but a  solution for custom blockers should still be a light lift solution and completely doable. It's essentially a mirror of of the OVH blockers, just one where we can specify the message and the detecting file path. 

----i--------e-- /var/cpanel/elevate-blockers.security-advisor
----i--------e-- /var/cpanel/elevate-blockers

the security advisor file, accept HTML in the msg while the other relies on \n

toddr:
Thank you for the request. We are loathed to tell someone with root that they cannot upgrade. A compromise we intend to implement is that if the file `/var/cpanel/elevate-noc-recommendations` is present when they have started their upgrade, it will emit the contents of the file to the console and give them a `proceed? (yes/no)`. 

Hopefully, this will provide mitigation for your concern?

jasonmpotter:
HI Todd, thanks for the response and the compromise. It's a good one for sure and I understand your position with denying someone with root from performing any operation. Unfortunately, server owners are granted root regardless of skill level due to needing to access WHM to manage the server. We offer fully managed cPanel servers so we are confident that customers will prefer our stance on non-in-place upgrades in favor of a dedicated migrations team. This is how we've always operated, so it won't be jarring to our customer base.

[...]

PR #201, opened 2023-03-01
Check for warning file from server provider

sloanebernstein:
If the file `/var/cpanel/elevate-noc-recommendations` exists, present the user with a generic warning asking them to check with their server provider. If the file has contents, display those instead of the generic warning. Prompt for an acknowledgement that the user has done what has been requested, and bail out if this is not answered in the positive.

Implements #198.


sloanebernstein:
How should we document this?

PR #205, opened 2023-03-08
Also update the kernel command line arguments in the GRUB environment file

sloanebernstein:
Kernels installed after ELevate has upgraded the system do not hard-code their command line arguments into the individual bootloader entry but instead refer to a variable in GRUB's persistent environment, stored in `/boot/grub2/grubenv`. Although a full `grub2-mkconfig` does appear to update the `kernelopts` GRUB environment variable from `GRUB_CMDLINE_LINUX` in the contents of `/etc/default/grub`, the only time this is run when a kernel is installed, it specifically excludes environment variable updates. The only other thing that explicitly touches the GRUB environment is to change the `saved_entry` variable to update the default boot option.

Since a full `grub2-mkconfig` is overkill just to update the environment, use the `grub2-editenv` utility to update the appropriate variable.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -216,7 +216,7 @@ use constant IGNORE_OUTDATED_SERVICES_FILE => q[/etc/cpanel/local/ignore_outdate
 use constant LEAPP_REPORT_JSON => q[/var/log/leapp/leapp-report.json];
 use constant LEAPP_REPORT_TXT  => q[/var/log/leapp/leapp-report.txt];
 
-use constant OVH_MONITORING_TOUCH_FILE => q[/var/cpanel/acknowledge_ovh_monitoring_for_elevate];
+use constant OVH_MONITORING_TOUCH_FILE      => q[/var/cpanel/acknowledge_ovh_monitoring_for_elevate];
 use constant NOC_RECOMMENDATIONS_TOUCH_FILE => q[/var/cpanel/elevate-noc-recommendations];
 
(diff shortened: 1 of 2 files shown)


atoomic (approved):

elevate-cpanel
```diff
@@ -1148,6 +1150,26 @@ sub check_and_fix_grub ($self) {
     return unless $grub_conf =~ s/GRUB_CMDLINE_LINUX="(.+?)"/GRUB_CMDLINE_LINUX="$1 net.ifnames=0"/m;
     File::Slurper::write_binary( DEFAULT_GRUB_FILE, $grub_conf );
 
+    # Also save net.ifnames=0 to the GRUB environment:
+    my $grubenv = Cpanel::SafeRun::Simple::saferunnoerror( GRUB_EDITENV, GRUB_ENV_FILE, 'list' );
+    foreach my $line ( split "\n", $grubenv ) {
+        my ( $name, $value ) = split "=", $line, 2;
+        next unless $name eq "kernelopts";
+
+        if ( $value !~ m/\bnet\.ifnames=0\b/a ) {
```
atoomic, line 1159: probably better to check for `net.ifnames=` and not the value?

sloanebernstein: > probably better to check for `net.ifnames=` and not the value?

probably

toddr: looks good except for this.

toddr: Let's remove `0\b` from the regex and I think we're good.

PR #206, opened 2023-03-09
Move the last chance warnings to before stage 1

sloanebernstein:
Currently, several warnings, including "your chance now to cancel the update process", take place within stage 1 proper. As such, by the time a user decides to cancel at the last moment, they are already committed to the upgrade. Because the dialog implemented in #201 also occurs right before this warning, replying in the negative also traps the user into the upgrade process.

The call chain used to arrive at the old location for the code is:

run_stage_1 <- _run_service <- run_service_and_notify <- start

The only other way the flow of execution converges on the same path to reach stage 1 is where the handler for `--service` invokes `run_service_and_notify` directly. Since that isn't an interactive process, there is no merit to presenting these warnings under those circumstances, which justifies moving these warnings into `start`.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -443,6 +443,8 @@ sub start ($self) {
 
     }
 
+    $self->give_last_chance();
+
     system touch => LOG_FILE;
 
     bump_stage();    # init stage number to 1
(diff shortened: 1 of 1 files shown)


elevate-cpanel
```diff
@@ -443,6 +443,8 @@ sub start ($self) {
 
     }
 
+    $self->give_last_chance();
```
atoomic, line 446: I understand that this is fixing a bug but this is now displaying that `last_chance` warning so early that we should not display it when running only the checks. 
Previously it was only happening when attempting an elevation.
It seems that it's going to always occur

atoomic (changes requested):
do not display last chance when only checking blockers

elevate-cpanel
atoomic: sorry this is probably good as this is called from start and check is another code path

atoomic (approved):

PR #219, opened 2023-03-27
Fix #151 and another bug I found along the way

### LINKED ISSUE #151: Do not complain about repo files when no packages from the repo are installed.
We want to change the blocker to block if rpms are installed from a repo we don't know about. 

We want to warn if repo files are present and not fully disabled when unknown.

troglodyne:

See second commit in set for "the bug I found along the way."

Internal tracking/testing notes can be found in BC-6268.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -2356,7 +2356,6 @@ sub yum_list ( $invalidate_cache = 0 ) {
     while ( my $line = shift @lines ) {
         last if $line && $line =~ m/^Installed Packages/;
     }
-    shift @lines;
 
     my %repos;
     while ( my $line = shift @lines ) {
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -232,43 +232,134 @@ use constant GRUB_EDITENV      => '/usr/bin/grub2-editenv';
 use constant GRUB_ENV_FILE     => '/boot/grub2/grubenv';
 
 use constant DISABLE_MYSQL_YUM_REPOS => qw{
-  Mysql57.repo
-  Mysql80.repo
-
-  MariaDB102.repo
-  MariaDB103.repo
-  MariaDB105.repo
-  MariaDB106.repo
-
-  mysql-community.repo
-
+  mysql-connectors-community
+  mysql-tools-community
+  mysql55-community
+  mysql56-community
+  mysql57-community
+  mysql80-community
+  mysql-tools-preview
+  mysql-cluster-7.5-community
+  mysql-cluster-7.6-community
+  mysql-connectors-community-source
+  mysql-tools-community-source
+  mysql55-community-source
+  mysql56-community-source
+  mysql57-community-source
+  mysql80-community-source
+  mysql-tools-preview-source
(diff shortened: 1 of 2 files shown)


elevate-cpanel
```diff
@@ -232,43 +232,134 @@ use constant GRUB_EDITENV      => '/usr/bin/grub2-editenv';
...
+  extras
+  centosplus
+  cr
+  base-debuginfo
+  c7-media
+  base-source
+  updates-source
+  extras-source
+  centosplus-source
+  C7.0.1406-base
```
toddr, line 291: Is this for cloudlinux? we don't support that yet.

troglodyne: That was just names of stuff I pulled from the centos-vault repo. No idea why those all looked so odd.

elevate-cpanel
```diff
@@ -232,43 +232,134 @@ use constant GRUB_EDITENV      => '/usr/bin/grub2-editenv';
...
-
+  mysql-connectors-community
+  mysql-tools-community
+  mysql55-community
+  mysql56-community
+  mysql57-community
+  mysql80-community
+  mysql-tools-preview
+  mysql-cluster-7.5-community
+  mysql-cluster-7.6-community
```
atoomic, line 243: maybe we should consider RegExp for some of these entries? which would provide an easier maintenance
example: `mysql-cluster-\d+\.\d+-community`

elevate-cpanel
```diff
@@ -232,43 +232,134 @@ use constant GRUB_EDITENV      => '/usr/bin/grub2-editenv';
...
+  C7.7.1908-base
+  C7.7.1908-updates
+  C7.7.1908-extras
+  C7.7.1908-centosplus
+  C7.7.1908-fasttrack
+  C7.8.2003-base
+  C7.8.2003-updates
+  C7.8.2003-extras
+  C7.8.2003-centosplus
+  C7.8.2003-fasttrack
```
atoomic, line 335: same here RegExp++? (apply to other names where it makes sense too)

elevate-cpanel
```diff
@@ -2265,7 +2356,6 @@ sub yum_list ( $invalidate_cache = 0 ) {
     while ( my $line = shift @lines ) {
         last if $line && $line =~ m/^Installed Packages/;
     }
-    shift @lines;
```
atoomic, line 2268: indeed not sure why that `shift` was there... the first line after `Installed Packages` seems a legit package

elevate-cpanel
```diff
@@ -232,43 +232,134 @@ use constant GRUB_EDITENV      => '/usr/bin/grub2-editenv';
...
-  kernelcare.repo
-  wp-toolkit-cpanel.repo
-  imunify360-ea-php-hardened.repo
-  droplet-agent.repo
+  base
+  updates
+  extras
+  centosplus
+  cr
+  base-debuginfo
```
toddr, line 285: no debuginfo

elevate-cpanel
toddr: Let's do this in a follow up commit

PR #226, opened 2023-03-29
Initial FatPack the elevate script

atoomic:
This is a start: we can then consider adding more as future iterations 
```
 lib/Elevate/Check.pm
 lib/Elevate/Cleanup.pm
 lib/Elevate/Component/Apache.pm
 lib/Elevate/Component/Imunify.pm
 lib/Elevate/Component/KernelCare.pm
...
 lib/Elevate/Notify.pm
```

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -2786,7 +2786,7 @@ sub remove_nixstats ($self) {
 
     my $service = q[nixstatsagent];
 
-    my $is_enabled = _is_service_enabled($service);
+    my $is_enabled = $self->service->is_enabled($service);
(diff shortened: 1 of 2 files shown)


maint/marker
troglodyne, line 1: Should this be gitignored instead if it is just generated?

atoomic: heredoc in Makefile is doable but require a special flag and was avoiding it 
maybe we can move that string to Makefile a blob

atoomic:
let's focus on #228 which is an improvement on that baseline

PR #245, opened 2023-04-28
Fix more bad ssystem calls

troglodyne:
In this case, $self already was an argument to the subroutine, so the replacement was rather straightforward.


script/elevate-cpanel.PL
```diff
@@ -1538,7 +1538,7 @@ sub remove_rpms_from_repos ( $self, @repo_list ) {
 
     INFO( "Removing packages for " . join( ", ", @repo_list ) );
 
-    __PACKAGE__->ssystem( qw{/usr/bin/yum -y erase}, @to_remove );
```
toddr, line 1541: Can we remove the dead subroutine?

PR #246, opened 2023-05-01
Replace last remaining instances of issues reference with cPSupport

troglodyne:
Pretty sure I got em all this time, nothing left that looks relevant anymore per git grep

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -831,8 +831,8 @@ EOS
 
         /scripts/elevate-cpanel --check
 
-        If your GRUB2 configuration has not been customized, consider reporting
-        this limitation to cPanel Support:
+        If your GRUB2 configuration has not been customized, you may want
+        to consider reaching out to cPanel Support for assistance:
         https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
         EOS
         }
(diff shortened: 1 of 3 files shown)


elevate-cpanel
```diff
@@ -832,7 +832,8 @@ EOS
         /scripts/elevate-cpanel --check
 
         If your GRUB2 configuration has not been customized, consider reporting
-        this limitation to https://github.com/cpanel/elevate/issues
+        this limitation to cPanel Support:
```
toddr, line 835: maybe not reporting? I can't find a good phrase :/

elevate-cpanel
```diff
@@ -1396,7 +1397,8 @@ EOS
             You should disable these repositories and remove packages installed from them
             before continuing the update.
 
-            Consider reporting this limitation to https://github.com/cpanel/elevate/issues
+            Consider reporting this limitation to cPanel Support:
```
toddr, line 1400: This should just come out.

Or maybe something like: If you need help removing the packages, support might be able to help... Not sure if this will end with damage and they'd be better off waiting until we support their repo.

troglodyne: Removing sounds good given the context. If they can run this on the CLI I would hope they know how to run yum or rpm

elevate-cpanel
troglodyne: "you may want to consider reaching out to cPanel Support for assistance" ?

PR #247, opened 2023-05-01
Block if the running and boot kernels differ

sloanebernstein:
The leapp framework requires the ability to instruct the boot loader to load the OS with a custom initrd. If this is not possible, the upgrade will not work. However, since we cannot directly determine whether we can affect the boot loader without rebooting[*], use kernel versions as a proxy.

[*] It is possible to guarantee agency over the boot loader by adding a dummy kernel parameter, which could be detected after reboot with `/proc/cmdline`; however, this method was considered to be overkill.


elevate-cpanel
```diff
@@ -216,6 +217,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
       JetBackup
       NICs
       EA4
+      BootKernel
```
toddr, line 220: This list indicates the order blockers will be checked BTW

sloanebernstein: Yes, and I (think I) want this blocker to be checked after package updates but before the Linode GRUB workaround. Is there another consideration that I might have missed?

PR #250, opened 2023-05-05
Add blocker for python36 and test

troglodyne:
Also fix explosion due to bad caller of Base in InvalidYumRepo check

A part of me says "a real shame we gotta do this" as it appears? `python3` is installed by default on c7. That said, only place where this could harm things is if for some reason a customer has cpanel-yarn/cpanel-node/etc. packages which are all generally development packages and not something customers would have installed (as they, oddly enough, depend on python3).

### DIFF
--- a/lib/Elevate/Blockers/Python.pm
+++ b/lib/Elevate/Blockers/Python.pm
@@ -20,10 +20,10 @@ sub check ($self) {
     my $pkg = Cpanel::Pkgr::what_provides('python36');
     return unless $pkg && Cpanel::Pkgr::is_installed($pkg);
     return $self->has_blocker( <<~"END" );
-    python36 packages have been detected as installed.
-    These can interfere with the elevation process.
-    Please remove these packages before elevation:
-    yum remove python36*
+    A package providing python36 has been detected as installed.
+    This can interfere with the elevation process.
+    Please remove it before elevation:
+    yum remove $pkg
     END
 }
 
--- a/lib/Elevate/Blockers/Python.pm
+++ b/lib/Elevate/Blockers/Python.pm
@@ -1,0 +1,30 @@
+package Elevate::Blockers::Python;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Blockers::Python
+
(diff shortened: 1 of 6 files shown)


troglodyne:
Parse error? bleh.

Will look at it in a few mins

atoomic:
note: test is failing due to incorrect usage of `Test::Cpanel::Policy`
```
Can't locate Test/Cpanel/Policy.pm in @INC
```

lib/Elevate/Blockers/Python.pm
```diff
@@ -0,0 +1,30 @@
...
+
+use parent qw{Elevate::Blockers::Base};
+
+use Cpanel::Pkgr  ();
+
+sub check ($self) {
+    my $pkg = Cpanel::Pkgr::what_provides('python36');
+    return unless $pkg && Cpanel::Pkgr::is_installed($pkg);
+    return $self->has_blocker( <<~"END" );
+    python36 packages have been detected as installed.
```
atoomic, line 23: not sure why we use plural here for `packages` as we only detected `one`?

atoomic (approved):

lib/Elevate/Blockers/Python.pm
atoomic: but yeap if `python36` is installed, very likely you have some other `python36-*` packages installed too :-)

troglodyne: Will push commit to try to address it

t/blocker-Python.t
```diff
@@ -0,0 +1,47 @@
...
+use Test2::Bundle::Extended;
+use Test2::Tools::Explain;
+use Test2::Plugin::NoWarnings;
+use Test::MockModule qw{strict};
+
+use Elevate::Blockers::Python ();
+
+{
+    package bogus::cpev;
+    sub _abort_on_first_blocker { return 0 }
```
atoomic, line 18: we should prefer using `cpev->new( _abort_on_first_blocker => 1 );`

atoomic (approved):

PR #251, opened 2023-05-05
Block on license issue

sloanebernstein:
Transient license issues may cause problems for the upgrade. Do a best-effort check to figure out if their license is okay, right this moment and block if that fails.

### DIFF
--- a/lib/Elevate/Blockers/WHM.pm
+++ b/lib/Elevate/Blockers/WHM.pm
@@ -17,6 +17,7 @@ use Elevate::Notify    ();
 
 use Cpanel::Version::Tiny ();
 use Cpanel::Update::Tiers ();
+use Cpanel::License       ();
 use Cpanel::Pkgr          ();
 
 use parent qw{Elevate::Blockers::Base};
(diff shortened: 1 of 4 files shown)


lib/Elevate/Blockers/WHM.pm
```diff
@@ -62,6 +64,44 @@ sub _blocker_is_newer_than_lts ($self) {
     return 0;
 }
 
+sub _blocker_cpanel_needs_license ($self) {
+    if ( !Cpanel::License::is_licensed( skip_max_user_check => 1 ) ) {
```
atoomic, line 68: style: probably easier to read like this 
```
return 1 if Cpanel::License::is_licensed( skip_max_user_check => 1 );
....
return 0;
```

lib/Elevate/Blockers/WHM.pm
```diff
@@ -62,6 +64,44 @@ sub _blocker_is_newer_than_lts ($self) {
     return 0;
 }
 
+sub _blocker_cpanel_needs_license ($self) {
```
atoomic, line 67: maybe consider adding some unit tests for that new function?

atoomic (approved):

lib/Elevate/Blockers/WHM.pm
sloanebernstein: Considering it. I don't want to drill down too far into cPanel internals though.

atoomic (approved):

PR #254, opened 2023-05-09
fix #170

### LINKED ISSUE #170: mysql config lost upon completion
I just realized today that after I migrated from CENTOS 7.9 to Alma my mariadb config file was replaced with the default config.
My mariadb version stayed the same, 10.5.

If this is intended to work this way - I don't remember a warning stating that this would happen. Some warnings would be nice as i would have definitely backed up my config file. transaction log settings were lost, query cache settings were lost, among many others.

If this is unintended - a my.cnf backup/restore should be executed. 

I don't see any reason why the config file should be replaced with the default, especially if the version is remaining the same.

troglodyne:

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -2931,9 +2931,7 @@ EOS
 
     use cPstrict;
 
-    use Elevate::Constants ();
-
-    use Cwd ();
+    use File::Copy ();
 
     # use Log::Log4perl qw(:easy);
     INIT { Log::Log4perl->import(qw{:easy}); }
@@ -2941,6 +2939,8 @@ EOS
     # use Elevate::Components::Base();
     our @ISA;
     BEGIN { push @ISA, qw(Elevate::Components::Base); }
+
+    my $cnf_file = '/etc/my.cnf';
 
     sub pre_leapp ($self) {
 
@@ -2965,6 +2965,8 @@ EOS
 
         cpev::update_stage_file( { 'mysql-version' => $mysql_version } );
 
+        File::Copy::copy( $cnf_file, "$cnf_file.rpmsave_pre_elevate" ) or WARN("Couldn't backup $cnf_file to $cnf_file.rpmsave_pre_elevate: $!");
+
         $self->_cleanup_mysql_57_packages();
         $self->_cleanup_mysql_80_packages();
         $self->_cleanup_mysql_102_packages();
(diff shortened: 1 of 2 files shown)


lib/Elevate/Components/MySQL.pm
```diff
@@ -65,6 +63,12 @@ sub _reinstall_mysql_packages {
 
     my ( $major, $minor ) = split( /\./, $mysql_version );
 
+    # Try to restore any .rpmsave'd configs before we reinstall
```
atoomic, line 66: should we consider any `.rpmsave` file? what if that `.rpmsave` is several months/years old? is it still a good idea?

troglodyne: Hmm, well in this case I very much doubt it will be anything *other* than the my.cnf that we just uninstalled earlier in the elevation process.

That said, it may be worth testing to put a broken one in place before elevation, then seeing if the broken one is restored (RPM in this case refusing to overwrite the old rpmsave'd one would be the only thing to lead to this). I'll leave a note to the tester about this.

atoomic: Alternatively we can check for the presence of the file pre-elevate, check it's timestamp and ignore the restore if this is the same one...

troglodyne:
Pushed up changes to use `.rpmsave_pre_elevate`, amended commit to be more descriptive since this is probably close to a final draft of this

elevate-cpanel
```diff
@@ -2965,6 +2965,8 @@ EOS
 
         cpev::update_stage_file( { 'mysql-version' => $mysql_version } );
 
+        File::Copy::copy( $cnf_file, "$cnf_file.rpmsave_pre_elevate" ) or WARN("Couldn't backup $cnf_file to $cnf_file.rpmsave_pre_elevate: $!");
```
toddr, line 2968: I think we should move not copy this so there's no confusion if the rpmsave file is about the recent RPM removal?

elevate-cpanel
```diff
@@ -2983,6 +2985,9 @@ EOS
 
         my ( $major, $minor ) = split( /\./, $mysql_version );
 
+        INFO("Restoring $cnf_file.rpmsave_pre_elevate to $cnf_file...");
```
toddr, line 2988: this should move the .rpmsave if the rpmsave was created as a part of the removal?

PR #261, opened 2023-05-11
Add logic to fix absolute symlinks in / and notice about us doing this

troglodyne:
Took more time than I thought because I was busy getting frustrated trying to test till I gave up on it

### DIFF
--- a/lib/Elevate/Blockers/AbsoluteSymlinks.pm
+++ b/lib/Elevate/Blockers/AbsoluteSymlinks.pm
@@ -1,0 +1,31 @@
+package Elevate::Blockers::AbsoluteSymlinks;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Blockers::AbsoluteSymlinks
+
+Blocker code to *warn* about absolute paths on symlinks in /, as we will be
+correcting these before run.
+
+=cut
+
+use cPstrict;
+
+use Elevate::Components::AbsoluteSymlinks ();
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Blockers::Base};
+
+sub check ($self) {
+    my %links = Elevate::Components::AbsoluteSymlinks::get_abs_symlinks();
+    WARN("Symlinks with absolute paths have been found in /:\n\t"
+        . join( ", ", sort keys(%links) )
+        ."This can cause problems during the leapp run, so\n"
+        .'these will be corrected to be relative symlinks before elevation.'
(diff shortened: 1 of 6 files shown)


lib/Elevate/Blockers/AbsoluteSymlinks.pm
```diff
@@ -0,0 +1,31 @@
...
+
+use Elevate::Components::AbsoluteSymlinks ();
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Blockers::Base};
+
+sub check ($self) {
+    my %links = Elevate::Components::AbsoluteSymlinks::get_abs_symlinks();
+    WARN("Symlinks with absolute paths have been found in /:\n\t"
+        . join( ", ", keys(%links) )
```
atoomic, line 24: `sort`?

lib/Elevate/Components/AbsoluteSymlinks.pm
```diff
@@ -0,0 +1,49 @@
...
+}
+
+sub pre_leapp {
+    my %links = get_abs_symlinks();
+    return unless %links;
+    my $chdir = Cpanel::Chdir->new("/");
+    foreach my $link (keys(%links)) {
+        my $updated = substr( $links{$link}, 1 );
+
+        my $rand_uid = Cpanel::UUID::random_uuid();
```
atoomic, line 38: note: even if the probability of collision is low, it can conflict...

toddr: Conflict with what? ideally nothing is happening on the system when they're upgrading?

troglodyne: yea... I realize nothing is ever truly "random" ever either... I'm just trying to *avoid* failures here when possible. Collision chances here should be so vanishingly small as to not ever be hit in the wild.

lib/Elevate/Blockers/AbsoluteSymlinks.pm
troglodyne: Sure, makes sense

lib/Elevate/Components/AbsoluteSymlinks.pm
toddr, line 38: check that the file is not there before you run symlink?

atoomic: that works

atoomic (approved):

PR #263, opened 2023-05-15
if we run check-upgrade to a specific flavor, suggest to run the correct command

bigio:

### DIFF
--- a/lib/Elevate/Blockers.pm
+++ b/lib/Elevate/Blockers.pm
@@ -91,11 +91,18 @@ sub check ($self) {    # do_check - main  entry point
 
     $self->save( $blocker_file, { 'blockers' => $self->{'blockers'} } );
 
+    my $flavor = $self->getopt('upgrade-to');
     if ($has_blockers) {
         WARN( <<~'EOS' );
         Please fix the detected issues before performing the elevation process.
         EOS
     }
+    elsif (defined $flavor) {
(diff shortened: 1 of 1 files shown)


lib/Elevate/Blockers.pm
```diff
@@ -91,11 +91,18 @@ sub check ($self) {    # do_check - main  entry point
 
     $self->save( $blocker_file, { 'blockers' => $self->{'blockers'} } );
 
+    my $flavor = $self->getopt('upgrade-to');
     if ($has_blockers) {
         WARN( <<~'EOS' );
         Please fix the detected issues before performing the elevation process.
         EOS
     }
+    elsif (defined $flavor) {
```
atoomic, line 100: looks a great idea, `length` is more accurate than defined, but this works

atoomic (approved):
thanks for submittint this idea

lib/Elevate/Blockers.pm
```diff
@@ -91,11 +91,18 @@ sub check ($self) {    # do_check - main  entry point
...
+        INFO( <<~'EOS' );
+        There is no known blockers to start the elevation process.
+        You can consider running:
+            /scripts/elevate-cpanel --start --upgrade-to=$flavor
+        EOS
```
atoomic, line 105: missing `}`

PR #268, opened 2023-05-17
Add the 4 newest imunify rollout 'slots' to vetted yum repos

troglodyne:


lib/Elevate/Blockers/Repositories.pm
```diff
@@ -93,6 +93,10 @@ use constant VETTED_YUM_REPO => qw{
   imunify360-rollout-2
   imunify360-rollout-3
   imunify360-rollout-4
+  imunify360-rollout-5
```
toddr, line 96: We probably need to enhance this array to support regexes. CloudLinux does the same stuff.

PR #271, opened 2023-05-18
Clear Cpanel::OS' $instance cache because it's definitely wrong on rocky

troglodyne:

Should fix false positive failures we've been seeing, it cleared it up for me

### DIFF
--- a/script/elevate-cpanel.PL
+++ b/script/elevate-cpanel.PL
@@ -954,6 +954,7 @@ Restore removed packages during the previous stage.
 
 sub run_stage_4 ($self) {
     Cpanel::OS::flush_disk_caches();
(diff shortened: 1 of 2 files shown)


script/elevate-cpanel.PL
```diff
@@ -954,6 +954,7 @@ Restore removed packages during the previous stage.
 
 sub run_stage_4 ($self) {
     Cpanel::OS::flush_disk_caches();
```
atoomic, line 956: wonder if `flush_disk_caches` should not imply `clear_cache` also 
but this LGTM

toddr: That would probably complicate cPanel tests

PR #289, opened 2023-07-03
Fix #287

### LINKED ISSUE #287: [BUG] Should check MySQL status before starting the upgrade
**Describe the bug**

When MySQL service is disabled on the server pre-elevate process. 
The elevate fails in stage 4 due to MySQL being disabled.

**Expected behavior**
We should check MySQL service status and when disabled, do not expect it to be enabled post-elevate.

**Additional context**
It seems that the ELevate process will fail in stage 4 if the MySQL service is disabled on the server. To clarify, the MySQL package upgrade itself does complete without issue, but the upgrade process fails as the MySQL service is not restarted at the end:
```
Restarting mysql service.
(XID 52zc2f) The "mysql" service is disabled.
Mysql did not start up after upgrade; there may be a problem in /etc/my.cnf. at /usr/local/cpanel/Whostmgr/Mysql/Upgrade.pm line 855.
```

This causes ELevate to fail.  To fix this, you have to enable MySQL:
```
whmapi1 configureservice service=mysql enabled=1 monitored=0
```
You can then proceed with --continue

troglodyne:
Since this is a bug in cPanel v110 or earlier, there's really no reasonable expectation to fix this anywhere but "in the mix" so just enable MySQL before performing the upgrade so that it does not fail. If it was disabled previously, re-disable it upon upgrade completion, however that winds up falling out.

### DIFF
--- a/lib/Elevate/Components/MySQL.pm
+++ b/lib/Elevate/Components/MySQL.pm
@@ -62,6 +62,7 @@ sub _cleanup_mysql_packages ($self) {
 sub _reinstall_mysql_packages {
 
     my $mysql_version = cpev::read_stage_file( 'mysql-version', '' ) or return;
+    my $enabled = cpev::read_stage_file( 'mysql-enabled', '' ) or return;
 
     INFO("Restoring MySQL $mysql_version");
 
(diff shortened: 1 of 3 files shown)


lib/Elevate/Components/MySQL.pm
```diff
@@ -72,6 +72,17 @@ sub _reinstall_mysql_packages {
     INFO("Restoring $cnf_file.rpmsave_pre_elevate to $cnf_file...");
     File::Copy::copy( "$cnf_file.rpmsave_pre_elevate", $cnf_file ) or WARN("Couldn't restore $cnf_file.rpmsave: $!");
 
+    # Apparently the APITool is in fact unable to figure out that --output=json means I want JSON in this context,
+    # so just load the modules directly and do what the API does to gather the data I want.
+    require Cpanel::Services::Enabled;
+    my $enabled = Cpanel::Services::Enabled::is_enabled('mysql');
```
atoomic, line 78: let's store that status to the elevate status file so we can reuse that state later when reinstalling/restarting MySQL and do not block on it when disabled

lib/Elevate/Components/MySQL.pm
```diff
@@ -88,7 +99,10 @@ sub _reinstall_mysql_packages {
         while (1) {
             $c   = ( $c + 1 ) % 10;
             $out = Cpanel::SafeRun::Simple::saferunnoerror( qw{/usr/local/cpanel/bin/whmapi1 background_mysql_upgrade_status }, "upgrade_id=$id" );
-            die qq[Failed to restore MySQL $mysql_version: cannot check upgrade_id=$id] if $?;
+            if($?) {
```
toddr, line 102: We don't want to die even if it fails if it was disabled to begin with.

lib/Elevate/Components/MySQL.pm
```diff
@@ -72,6 +72,17 @@ sub _reinstall_mysql_packages {
...
+    # so just load the modules directly and do what the API does to gather the data I want.
+    require Cpanel::Services::Enabled;
+    my $enabled = Cpanel::Services::Enabled::is_enabled('mysql');
+    if( !$enabled ) {
+        INFO("MySQL is not enabled. This will cause the MySQL upgrade tool to fail. Temporarily enabling it to ensure the upgrade succeeds.");
+        Cpanel::SafeRun::Simple::saferunnoerror( qw{/usr/local/cpanel/bin/whmapi1 configureservice service=mysql enabled=1} );
+        # Pray it goes ok, as what exactly do you want me to do if this reports failure? May as well just move forward in this case without checking.
+
+    }
+
```
atoomic, line 85: what if the elevate process crash and we have to run `--continue` from there, we will use an incorrect state

atoomic (changes requested):
let's capture state or use a blocker?

troglodyne:
Pushed changes, I think they should address the concerns

PR #292, opened 2023-07-05
Allow upgrade to proceed with digitalocean-agent

sloanebernstein:
The `digitalocean-agent` repo with its `do-agent` package works exactly the same way as the `droplet-agent` repo with its `droplet-agent` package, so perform the same steps done with the latter (reinstall the package if it and its repo are present) to the former.


sloanebernstein:
The re-installation of the DO packages has been removed, as it seems to be unnecessary.

atoomic (approved):

toddr (approved):
less code++

PR #295, opened 2023-08-07
Vet elasticsearch-7.x repoid

sloanebernstein:
The officially recommended repo file for ElasticSearch uses `elasticsearch` as the YUM repo ID; however, a third-party document seems to use `elasticsearch-7.x` for whatever reason. The contents of the file are more or less equivalent.


sloanebernstein:
It was requested to document the reasons new repos are placed into the vetted list. However, the list is a `qw{}` construct, so in-line comments won't work. If we really want a block of comments above the existing list with this info, I can add that.

PR #312, opened 2023-09-07
Mysql cover

atoomic:

### DIFF
--- a/lib/Elevate/Components/MySQL.pm
+++ b/lib/Elevate/Components/MySQL.pm
@@ -18,6 +18,7 @@ use Log::Log4perl qw(:easy);
 use parent qw{Elevate::Components::Base};
 
 my $cnf_file = '/etc/my.cnf';
+
 sub pre_leapp ($self) {
 
     $self->run_once("_cleanup_mysql_packages");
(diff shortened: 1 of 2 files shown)


lib/Elevate/Components/MySQL.pm
```diff
@@ -110,19 +114,30 @@ sub _reinstall_mysql_packages {
...
+        Cpanel::SafeRun::Simple::saferunnoerror(qw{/usr/local/cpanel/bin/whmapi1 configureservice service=mysql enabled=0}) if !$enabled;
 
         if ( $status eq 'success' ) {
             INFO("MySQL $mysql_version restored");
         }
         else {
-            FATAL("Failed to restore MySQL $mysql_version: upgrade $id status '$status'");
+            my $msg = "Failed to restore MySQL $mysql_version: upgrade $id status '$status'";
+
+            FATAL($msg);
```
toddr, line 136: does FATAL die?

atoomic: if it was the case that code would already be broken...

atoomic: `LOGDIE` is the one dying

PR #319, opened 2023-10-30
Allow user to customize LEAPP_OVL_SIZE

### LINKED ISSUE #318: [BUG]  LEAPP_OVL_SIZE should be able to be changed
**Describe the bug**
The script [sets the value of LEAPP_OVL_SIZE](https://github.com/cpanel/elevate/blob/0abeaa36acb021f64e9964a4740821a9c831101e/script/elevate-cpanel.PL#L1416) before calling leapp without checking if the admin tried to set it to something else. This means if that size isn't big enough, there is no easy way for the admin to override it.

**To Reproduce**
1. Try to upgrade a machine where 3000 isn't big enough.
2. Set the environment variable `LEAPP_OVL_SIZE=200000` and see that the actual overlay size doesn't change

**Expected behavior**
The admin should be able to increase the size if needed.

atoomic:
Fix #318

If the LEAPP_OVL_SIZE is set capture the value
at startup and honor it later.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -5189,6 +5189,10 @@ sub start ($self) {
         update_stage_file( { manual_reboots => 1 } );
     }
 
+    if ( my $ovl_size = $ENV{LEAPP_OVL_SIZE} ) {
+        update_stage_file( { env => { LEAPP_OVL_SIZE => $ovl_size } } );
+    }
+
     if ( $self->getopt('no-leapp') ) {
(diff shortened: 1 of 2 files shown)


atoomic:
note: this is currently in start but should also probably do the same when running `--continue` so we can tweak it on demand later

atoomic:
idea: capture env variable at startup and continue so the service can consume the variables

PR #340, opened 2024-01-09
Teach should_run_leapp to check for the CLI option

### LINKED ISSUE #332: Blocker check for containers should not run when --no-leap is passed

cPholloway:
Previously, the check used by blockers to determine if the script will run leapp was incompatible with the check option as it only read a file that was written just prior to running stage 1.  This updates the check to see if we should run leapp to also check to see if the --no-leapp option was passed via CLI in order to make the sub compatible with the check option.

Fixes #332

### DIFF
--- a/script/elevate-cpanel.PL
+++ b/script/elevate-cpanel.PL
@@ -1696,7 +1696,9 @@ sub setup_answer_file {
 sub should_run_leapp ($self) {
 
     # we store the no_leapp option, but prefer using a positive check instead
-    my $no_leapp = read_stage_file( 'no_leapp', 0 );
+    # we need to check to see if the no-leapp option is passed via CLI here too in order
+    # to allow users to run this script with the '--check --no-leapp' options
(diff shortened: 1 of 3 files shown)


script/elevate-cpanel.PL
```diff
@@ -1696,7 +1696,9 @@ sub setup_answer_file {
 sub should_run_leapp ($self) {
 
     # we store the no_leapp option, but prefer using a positive check instead
-    my $no_leapp = read_stage_file( 'no_leapp', 0 );
+    # we need to check to see if the no-leapp option is passed via CLI here too in order
+    # to allow users to run this script with the '--check --no-leapp' options
+    my $no_leapp = read_stage_file( 'no_leapp', 0 ) || $self->getopt('no-leapp');
```
toddr, line 1701: should we be checking the getopt first prior to going out to disk? I'm wondering if there are gotchas going in this order.

PR #343, opened 2024-01-10
Add blocker if cPanel is not running version 110

### LINKED ISSUE #337: Add blocker if cPanel is not running version 110 (last version to support C7)
Right now it checks that the latest version is installed of any tier. If it doesn't already it should check that the latest version of ONLY the 110 tier is installed.

cPholloway:
This is the last version to support c7, so we want to fix this script to only run on this version.  This will minimize testing and risk moving forward.

Fixes #337

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -15,7 +15,7 @@ jobs:
     env:
       PERL_USE_UNSAFE_INC: 1
       CPANEL_BIN_PATH: /usr/local/cpanel/3rdparty/bin
-      CPANEL_PERL: /usr/local/cpanel/3rdparty/perl/532/bin/perl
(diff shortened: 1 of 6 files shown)


toddr:
Please update the changelog file with a line indicating what happened in this change.

cPholloway:
This is currently blocked by the CI we have in place.  The container that the CI uses is based on 102/Perl 5.32.0.  We will need to update the CI to use a container based on 110/Perl 5.36.0.  Unfortunately, that container does not yet exist so we will need to build it in the container project that we use:

https://github.com/cpanel/docker-perl-compiler/tree/perl-v5.32.0

I am going to move on to 336 for now and revisit how to make the new container once 336 is fixed.

atoomic (approved):

PR #353, opened 2024-01-17
Improve warning before starting the upgrade process

### LINKED ISSUE #338: Enhance warning during: "This is your chance now to cancel the update process"
We advise/mention that customers should have a full server backup multiple times in the docs, but to be safer, we should consider having them acknowledge this in the script itself.

NOTE:  This option should have a --confirm flag or a --yes flag to allow customers to avoid STDIN

timmullin:
Fixes #338


elevate-cpanel
```diff
@@ -5273,12 +5285,32 @@ EOS
...
+        if (
+            !IO::Prompt::prompt(
+                '-one_char',
+                '-yes_no',
+                '-tty',
+                -default => 'n',
+                "Do you wish to proceed with the upgrade process [y/N]: ",
+            )
+        ) {
+            INFO("The user has chosen not to perform the upgrade");
```
toddr, line 5307: I'm not sure about the language on this sentence.

toddr: specifically "the user". Maybe just "Upgrade aborted"?

script/elevate-cpanel.PL
```diff
@@ -609,12 +620,30 @@ EOS
     say <<EOS;
 This is your chance now to cancel the update process.
 You can hit CTRL-C to abort, otherwise the install starts in a few seconds.
```
cPholloway, line 622: Lines 621 and 622 no longer make sense after this update

script/elevate-cpanel.PL
```diff
@@ -609,12 +620,30 @@ EOS
...
+
+It is *highly* recommended that you have a full backup or snapshot
+of your server before proceeding.
 EOS
 
     do { sleep 1; print '.' }
       for 1 .. 6;
     print "\n";
```
cPholloway, line 630: Lines 628-630 should also be removed since we require the user to confirm now (unless they give the option to bypass the confirmation.)

script/elevate-cpanel.PL
annes449: If we remove 621 and 622, and, 628-630, then, won't 624, and, 625 scroll right past them?

script/elevate-cpanel.PL
```diff
@@ -609,12 +620,30 @@ EOS
...
 
+    if ( !$self->getopt('yes') && !$self->getopt('non-interactive') ) {
+        if ( !IO::Prompt::prompt(
+            '-one_char',
+            '-yes_no',
+            '-tty',
+            -default => 'n',
+            "Do you wish to proceed with the upgrade process [y/N]: ",
+        ) ) {
+            INFO("The user has chosen not to perform the upgrade");
```
annes449, line 640: I think we could just exit without INFO here.
For example:
```
# dnf update
-snip-
Total download size: 579 M
Is this ok [y/N]: N
# 
```

elevate-cpanel
annes449: I think we could just exit without INFO here.
For example:
```
# dnf update
-snip-
Total download size: 579 M
Is this ok [y/N]: 
# 
```

timmullin: I was thinking it might be good to have something in the log that showed that the upgrade had been deliberately canceled.  
Maybe something like "The upgrade has been canceled"?

timmullin: Yeah, I can see that "The user has chosen not to perform the upgrade" might seem a more awkward way of saying it.

script/elevate-cpanel.PL
timmullin: Makes sense.  Having a prompt eliminates the need for the six second delay.

annes449: Could we move or copy 624-625 to the --help output?

timmullin: I imagine there would be no problem copying this over to --help.  I think we still want it here has well so they have the warning before answering the prompt.

elevate-cpanel
```diff
@@ -4714,6 +4715,7 @@ Helps to upgrade CentOS 7 cPanel servers to AlmaLinux 8 or Rocky Linux 8.
        --clean                        Cleanup scripts and files created by elevate-cpanel
        --upgrade-to=[rocky|almalinux] Update to AlmaLinux 8 or Rocky Linux 8
                                       [default to 'almalinux']
+       --yes, --non-interactive       Skip the yes/no prompt before proceeding with the upgrade.
```
annes449, line 4718: Could we add a '[default to 'no']' here?

toddr: I'm also not sure if we need --non-interactive. That was just an idea. Do we really want it?

timmullin: I'm not sure if we need --non-interactive.  Probably just having --yes will suffice.

cPholloway:
@timmullin 

The production code part of this commit looks good, but I did have two questions:

1. Should we add a new unit test for this change?
1. Should we add a new changelog entry for this change?

cPholloway (changes requested):
Marking request changes for now as this is currently waiting on a question to be answered.

annes449 (approved):
QA Testing is all good.

cPholloway (approved):

PR #357, opened 2024-01-19
Provide expected version when cPanel version is not up to date.

### LINKED ISSUE #339: Provide helpful information when a 'cPanel $version does not appear to be up to date' Blocker is encountered
- **Provide helpful information when a 'cPanel $version does not appear to be up to date' Blocker is encountered**

- This is the Blocker the client sees when their server is not running the most recent version of cPanel v110:

```
......
* 2024-01-08 21:43:45 [INFO] Successfully verified signature for cpanel (key types: release, development).
* 2024-01-08 21:43:45 [ERROR] This installation of cPanel (11.110.0.18) does not appear to be up to date.
  Please upgrade cPanel to a most recent version.
* 2024-01-08 21:43:45 [WARN] Please fix the detected issues before performing the elevation process.
  Read More: https://cpanel.github.io/elevate/blockers/
```

- For v102 the information is the same:

```
......
info [elevate-cpanel] Successfully verified signature for cpanel (key types: release, development).
* 2024-01-08 22:55:59 [ERROR] This installation of cPanel (11.102.0.36) does not appear to be up to date.
  Please upgrade cPanel to a most recent version.
* 2024-01-08 22:55:59 [WARN] Please fix the detected issues before performing the elevation process.
  Read More: https://cpanel.github.io/elevate/blockers/

```

- Maybe, we could tell them what version we are looking for, suggest they make certain updates are enabled in `/etc/cpupdate.conf` and provide a command to perform the upcp, like, If updates are not disabled on your server, then, you may execute the update now using `/usr/local/cpanel/scripts/fix-cpanel-perl; /usr/local/cpanel/scripts/upcp --cron`

timmullin:
Fixes #339

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -2026,13 +2026,12 @@ EOS
 
     sub _blocker_cpanel_needs_update ($self) {
         if ( !$self->getopt('skip-cpanel-version-check') ) {
-            my $tiers_obj = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
-            if ( !grep { Cpanel::Version::Compare::compare( $Cpanel::Version::Tiny::VERSION_BUILD, '==', $_ ) } $tiers_obj->get_flattened_hash()->@{qw/edge current release stable lts/} ) {
-                my $hint = '';
-                $hint = q[hint: You can skip this check using --skip-cpanel-version-check] if $Cpanel::Version::Tiny::VERSION_BUILD =~ 9999;
+            my $tiers_obj        = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
+            my $expected_version = $tiers_obj->get_flattened_hash()->{'11.110'};
+            if ( !Cpanel::Version::Compare::compare( $Cpanel::Version::Tiny::VERSION_BUILD, '==', $expected_version ) ) {
                 return $self->has_blocker( <<~"EOS" );
             This installation of cPanel ($Cpanel::Version::Tiny::VERSION_BUILD) does not appear to be up to date.
-            Please upgrade cPanel to a more recent version. $hint
+            Please upgrade cPanel to $expected_version.
             EOS
             }
         }
--- a/lib/Elevate/Blockers/WHM.pm
+++ b/lib/Elevate/Blockers/WHM.pm
@@ -107,13 +107,12 @@ sub _blocker_cpanel_needs_license ($self) {
 
 sub _blocker_cpanel_needs_update ($self) {
     if ( !$self->getopt('skip-cpanel-version-check') ) {
-        my $tiers_obj = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
-        if ( !grep { Cpanel::Version::Compare::compare( $Cpanel::Version::Tiny::VERSION_BUILD, '==', $_ ) } $tiers_obj->get_flattened_hash()->@{qw/edge current release stable lts/} ) {
-            my $hint = '';
-            $hint = q[hint: You can skip this check using --skip-cpanel-version-check] if $Cpanel::Version::Tiny::VERSION_BUILD =~ 9999;
+        my $tiers_obj        = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
+        my $expected_version = $tiers_obj->get_flattened_hash()->{'11.110'};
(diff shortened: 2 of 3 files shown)


elevate-cpanel
```diff
@@ -2026,13 +2026,15 @@ EOS
 
     sub _blocker_cpanel_needs_update ($self) {
         if ( !$self->getopt('skip-cpanel-version-check') ) {
-            my $tiers_obj = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
-            if ( !grep { Cpanel::Version::Compare::compare( $Cpanel::Version::Tiny::VERSION_BUILD, '==', $_ ) } $tiers_obj->get_flattened_hash()->@{qw/edge current release stable lts/} ) {
+            my $tiers_obj        = Cpanel::Update::Tiers->new( logger => Log::Log4perl->get_logger(__PACKAGE__) );
+            my $expected_version = $tiers_obj->get_flattened_hash()->{'lts'};
```
toddr, line 2030: This breaks in July when lts bumps to 118. Can we make it 11.110 instead @timmullin ?

elevate-cpanel
```diff
@@ -2026,13 +2026,15 @@ EOS
...
+            if ( !Cpanel::Version::Compare::compare( $Cpanel::Version::Tiny::VERSION_BUILD, '==', $expected_version ) ) {
                 my $hint = '';
                 $hint = q[hint: You can skip this check using --skip-cpanel-version-check] if $Cpanel::Version::Tiny::VERSION_BUILD =~ 9999;
                 return $self->has_blocker( <<~"EOS" );
             This installation of cPanel ($Cpanel::Version::Tiny::VERSION_BUILD) does not appear to be up to date.
-            Please upgrade cPanel to a more recent version. $hint
+            Please upgrade cPanel to $expected_version.
+            $hint
```
toddr, line 2037: not sure if we should be including $hint at this point :/

timmullin: Should I remove it?

toddr (approved):

toddr (changes requested):
Need improvements to remove the begin block check for older versions than 110

timmullin:
@toddr - I wonder if that check even belongs in the BEGIN block.  I'm thinking it would make sense to also run that check when they pass in --version.  For things like --start and --check they will already see it as a blocker, so the BEGIN check wouldn't be necessary.  And, for other things like --continue, having that test in the BEGIN block doesn't seem to have any purpose.

toddr:
> @toddr - I wonder if that check even belongs in the BEGIN block. I'm thinking it would make sense to also run that check when they pass in --version. For things like --start and --check they will already see it as a blocker, so the BEGIN check wouldn't be necessary. And, for other things like --continue, having that test in the BEGIN block doesn't seem to have any purpose.

The first check is designed to break if they're not on a cPanel system. We don't even think the script will compile in that case. You think we should take it out?

cPholloway:
@toddr @timmullin 

I think this change is fine as is.  Removing the cPanel version check from the begin block is what RE-56 is about.

timmullin:
> > @toddr - I wonder if that check even belongs in the BEGIN block. I'm thinking it would make sense to also run that check when they pass in --version. For things like --start and --check they will already see it as a blocker, so the BEGIN check wouldn't be necessary. And, for other things like --continue, having that test in the BEGIN block doesn't seem to have any purpose.
> 
> The first check is designed to break if they're not on a cPanel system. We don't even think the script will compile in that case. We want it to give a sane answer before it fails to load all the modules. You think we should take it out?

I think leaving the first check in the BEGIN block is just fine; the script shouldn't work on non-cPanel system.  It is the second check, checking the major version, that I don't think we need there since we check for that before attempting the upgrade.

cPholloway:
> > > @toddr - I wonder if that check even belongs in the BEGIN block. I'm thinking it would make sense to also run that check when they pass in --version. For things like --start and --check they will already see it as a blocker, so the BEGIN check wouldn't be necessary. And, for other things like --continue, having that test in the BEGIN block doesn't seem to have any purpose.
> > 
> > 
> > The first check is designed to break if they're not on a cPanel system. We don't even think the script will compile in that case. We want it to give a sane answer before it fails to load all the modules. You think we should take it out?
> 
> I think leaving the first check in the BEGIN block is just fine; the script shouldn't work on non-cPanel system. It is the second check, checking the major version, that I don't think we need there since we check for that before attempting the upgrade.

Removing the second check is the ask in RE-56.

timmullin:
So, if the issue with the second check is to be addressed in RE-56, then I guess this pull request should move forward?

PR #362, opened 2024-01-23
Add blocker for start mode if backup or upcp is currently executing

cPholloway:
Case RE-71: elevate-cpanel adds a motd notice that elevate is in progress and warns the sysadmin not make any changes while it is executing.  It also expects to reboot the system multiple times.  As such, we need to check to ensure that scripts/upcp and bin/backup are not executing before starting the elevate process.  scripts/upcp is especially problematic if it is running simultaneously as the reboot could leave the system in a very difficult to repair broken state.

Changelog: Add blocker for start mode if bin/backup or scripts/upcp is
 currently executing


timmullin (approved):

lib/Elevate/Blockers.pm
```diff
@@ -146,6 +146,16 @@ sub is_check_mode ($) {
     return $_CHECK_MODE;
 }
 
+sub set_abort_on_first_blocker ( $self, $value ) {
```
toddr, line 149: This probably just needs to be an accessor?

toddr: use Simple::Accessor qw(set_abort_on_first_blocker).

Then just update the things that check `_abort_on_first_blocker`

toddr (changes requested):

lib/Elevate/Blockers.pm
cPholloway: Unfortunately, we need the setter until we can do the refactor via RE-128.

lib/Elevate/Blockers.pm
```diff
@@ -119,8 +119,8 @@ sub _has_blockers ( $self, $check_mode = 0 ) {
         return 666;
     }
 
-    $_CHECK_MODE = !!$check_mode;                              # running with --check
-    $self->cpev->{_abort_on_first_blocker} = !$_CHECK_MODE;    # abort on first blocker
+    $_CHECK_MODE = !!$check_mode;    # running with --check
+    $self->set_abort_on_first_blocker( !$_CHECK_MODE );
```
toddr, line 123: $self->abort_on_first_blocker( !$_CHECK_MODE );

toddr (approved):

annes449 (approved):
QA is all good.

PR #365, opened 2024-01-29
Restore mongodb auth file after JetBackup restore

toddr:
Case RE-135

Changelog: Fix mongodb auth file after JetBackup restore

### DIFF
--- a/lib/Elevate/Components/Base.pm
+++ b/lib/Elevate/Components/Base.pm
@@ -63,4 +63,10 @@ sub run_once ( $self, $subname ) {
     return $run_once->( $cpev, $label, $code );
 }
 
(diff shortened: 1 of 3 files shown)


lib/Elevate/Components/Base.pm
```diff
@@ -63,4 +63,10 @@ sub run_once ( $self, $subname ) {
     return $run_once->( $cpev, $label, $code );
 }
 
+sub restore_rpmsave ( $self, $file ) {
```
toddr, line 66: This sub is here until Travis completes his work in a competing PR.

toddr:
Waiting for feedback from JetBackup before we make this change.

toddr:
JetBackup will fix this in their RPM.

PR #366, opened 2024-01-29
Remove cPanel version check in the BEGIN block

### LINKED ISSUE #347: `/scripts/elevate-cpanel --check` should check for fatal errors before checking for issues the client can resolve
- **`/scripts/elevate-cpanel --check` should check for fatal errors before checking for issues the client can resolve**

- When the client runs `/scripts/elevate-cpanel --check` on a v100.0.12 binary build running on CloudLinux 6.10, they are told the following:

```
# /scripts/elevate-cpanel --check
You need to upgrade your cPanel server to version 102or later before running this script.
```

- But, if the client updates the cPanel version to v102.0.36. and, again runs `/scripts/elevate-cpanel --check`, the following is displayed:

```
# /scripts/elevate-cpanel --check
Use of uninitialized value in split at /usr/local/cpanel/Cpanel/RestartSrv/Systemd.pm line 32.
* 2024-01-11 19:17:20 [FATAL] This script is designed to only run on CentOS 7 servers.
```

- **_We probably should have told the client that ELevate wasn't going to work on this server before giving them instructions on how to make it work._**

timmullin:
Case RE-56:

Fixes #347. Invoking the script with --check (or --start) will run a more thorough and informative version check as well as checking for any other issues of concern.

Changelog: Removed the simple version check from the BEGIN
  block of the script since this happens during the check
  operation.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -4936,13 +4936,6 @@ BEGIN {
         warn(qq[This script is designed to only run on cPanel servers.\n]);
         exit 1;
     };
-
-    $Cpanel::Version::Tiny::major_version >= Elevate::Constants::MINIMUM_LTS_SUPPORTED
-      or do {
-        warn qq[You need to upgrade your cPanel server to version ] . Elevate::Constants::MINIMUM_LTS_SUPPORTED    #
-          . qq[ or later before running this script.\n];
-        exit 1;
-      };
 }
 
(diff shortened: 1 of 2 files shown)


cPholloway (approved):

annes449 (approved):
QA testing passed.

PR #367, opened 2024-01-29
Restore config files for packages provided by the EA4 repo

cPholloway:
Case RE-38: Previously, we did restore any custom configuration files for packages provided by EA4.  This could lead to things like the PHP memory_limit being reset to default and custom modsec rules no longer being applied after the server was elevated.  This change restores custom configuration files for packages provided by the EA4 repo so that customizations will stick after the server has been elevated.

Changelog: Restore config files for packages provided by the EA4 repo

### DIFF
--- a/lib/Elevate/Components/EA4.pm
+++ b/lib/Elevate/Components/EA4.pm
@@ -13,6 +13,7 @@ Perform am EA4 backup pre-elevate then restore it after the elevation process.
 use cPstrict;
 
 use Elevate::Constants ();
+use Elevate::RPM       ();
 
 use Cwd           ();
 use Log::Log4perl qw(:easy);
@@ -42,6 +43,7 @@ sub pre_leapp ($self) {    # run to perform the backup
 
     $self->run_once('_backup_ea4_profile');
     $self->run_once('_backup_ea_addons');
+    $self->run_once('_backup_config_files');
     $self->run_once('_cleanup_rpm_db');
 
     return;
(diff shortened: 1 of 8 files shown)


lib/Elevate/Components/EA4.pm
```diff
@@ -52,6 +54,19 @@ sub post_leapp ($self) {
...
+    # On a new install, the RPM behavior for %config is to move the preexisting config file
+    # to '.rpmorig' and replace the config file with config file provided by the RPM
+    #
+    # On a new install, the RPM behavior for %config(noreplace) is to remove the preexisting
+    # config file and place the config file provided by the RPM at '.rpmnew'
+    #
+    # There should not be any need to restart services to pick up the new config files since the last
+    # step of stage 5 is to reboot the server so the services will be restarted and pick up the configs
+    # after this anyway
+    $self->run_once('_restore_config_files');
```
toddr, line 68: do we need to restart?

timmullin (approved):

lib/Elevate/RPM.pm
```diff
@@ -0,0 +1,84 @@
...
+    my $suffix = '.rpmsave';
+
+    foreach my $file (@files) {
+        next unless length $file;
+
+        my $backup_file = $file . $suffix;
+
+        next unless -e $backup_file;
+
+        File::Copy::move( $backup_file, $file ) or WARN("Unable to restore config file $backup_file: $!");
```
toddr, line 78: why not rename? This doesn't need to be portable to windows.

lib/Elevate/RPM.pm
```diff
@@ -0,0 +1,84 @@
...
+
+Logic wrapping the 'rpm' system binary
+
+=cut
+
+use cPstrict;
+
+use Log::Log4perl qw(:easy);
+
+use    # hide
```
toddr, line 17: hide from what?

toddr (changes requested):
Just need some feedback on some of  my questions.

lib/Elevate/RPM.pm
cPholloway: I used 'File::Copy' because elevate is already using the module in several different places.  If it was not already included, then I would have just used rename to avoid bringing in a new module.

lib/Elevate/RPM.pm
cPholloway: I'm not actually sure here.  This file was a copy/paste of 'lib/Elevate/Service.pm' and this is how that file had it.  Not a big deal to change it.

toddr: let's take it out. and even remove it from the other ones. I see no value and I think it's just going to cause cargo copying.

cPholloway: done

annes449 (approved):
QA is all good.

toddr (approved):

PR #375, opened 2024-02-16
Fix PostgreSQL user database check when feature disabled

timmullin:
Case RE-188:

For each user, check if it has the feature enabled before listing the databases for that user.

Changelog: Fixed PostgreSQL user database check for users
  that have the feature disabled.

### DIFF
--- a/lib/Elevate/Blockers/Databases.pm
+++ b/lib/Elevate/Blockers/Databases.pm
@@ -99,6 +99,9 @@ sub _has_mapped_postgresql_dbs ($self) {
     # Compile a list of users with PgSQL DBs:
     my @users_with_dbs;
     foreach my $user (@users) {
+        my $feature_json   = Cpanel::SafeRun::Simple::saferunnoerror( "/usr/local/cpanel/bin/uapi", "--user=$user", qw(--output=json Features has_feature name=postgres) );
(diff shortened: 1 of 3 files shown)


lib/Elevate/Blockers/Databases.pm
```diff
@@ -99,6 +99,9 @@ sub _has_mapped_postgresql_dbs ($self) {
     # Compile a list of users with PgSQL DBs:
     my @users_with_dbs;
     foreach my $user (@users) {
+        my $feature_json   = Cpanel::SafeRun::Simple::saferunnoerror( "/usr/local/cpanel/bin/uapi", "--user=$user", qw(--output=json Features has_feature name=postgres) );
+        my $feature_parsed = Cpanel::JSON::Load($feature_json);
```
toddr, line 103: does `Load` die if it cannot parse the JSON?

toddr (changes requested):
Questions.

lib/Elevate/Blockers/Databases.pm
timmullin: Yes.

timmullin:
We are going to try a different approach to handle this.

annes449 (commented):
QA is all good

PR #376, opened 2024-02-20
Add blocker to run "leapp preupgrade"

timmullin:
Case RE-90:

Add a blocker which runs "leapp preupgrade" to check for any issues that will prevent an upgrade.

Changelog: Add a pre-flight leapp check when checking for
  upgrade blockers.


lib/Elevate/Blockers/Leapp.pm
```diff
@@ -0,0 +1,50 @@
...
+use Cwd           ();
+use Log::Log4perl qw(:easy);
+
+sub check ($self) {
+
+    return if $self->is_check_mode();    # skip for --check
+
+    return unless $self->should_run_leapp;    # skip when --no-leapp is provided
+
+    $self->cpev->leapp->install(0);
```
toddr, line 28: just curious... what's (0)?

lib/Elevate/Leapp.pm
```diff
@@ -38,7 +38,7 @@ sub _build_yum ($self) {
     return Elevate::YUM->new( cpev => $self->cpev() );
 }
 
-sub install ($self) {
+sub install ( $self, $remove_kernel_devel = 1 ) {
```
toddr, line 41: This seems like a red flag. Why are we doing component work here. Could it be easily moved?

lib/Elevate/Leapp.pm
```diff
@@ -87,6 +96,64 @@ sub upgrade ($self) {
...
+
+    my @blockers;
+    my @ignored_blockers = qw(
+      check_installed_devel_kernels
+      verify_check_results
+    );
+    my $leapp_json_report = LEAPP_REPORT_JSON;
+
+    if ( !-e $leapp_json_report ) {
+        ERROR("Leapp did not generated the expected report file: $leapp_json_report");
```
toddr, line 109: bad english?

lib/Elevate/Leapp.pm
```diff
@@ -87,6 +96,64 @@ sub upgrade ($self) {
...
+        return [];
+    }
+
+    my $report = eval { Cpanel::JSON::LoadFile($leapp_json_report) } // {};
+    if ( my $exception = $@ ) {
+        ERROR("Unable to parse leapp report file ($leapp_json_report): $exception");
```
toddr, line 115: should this be a blocker if we can't parse it?

lib/Elevate/Leapp.pm
```diff
@@ -87,6 +96,64 @@ sub upgrade ($self) {
     return;
 }
 
+sub search_report_file_for_blockers ($self) {
+
+    my @blockers;
+    my @ignored_blockers = qw(
```
toddr, line 102: constant at top of file?

lib/Elevate/Leapp.pm
toddr: maybe return ['unable to parse leapp file thingy?']

toddr (changes requested):

lib/Elevate/Leapp.pm
timmullin: Yeah, we should move the kernel-devel uninstall to the code that calls install() when running the leapp upgrade.

lib/Elevate/Leapp.pm
timmullin: Typo.  My bad.  I'll s/generated/generate/

lib/Elevate/Leapp.pm
timmullin: I'm not sure if that should be a blocker or not.  Maybe it should.  If running "leapp preupgrade" generated an invalid report file, then something is very wrong.  If it can't run "leapp preupgrade" properly, then I wouldn't trust it to run "leapp upgrade".

lib/Elevate/Leapp.pm
timmullin: Since it is only used in this function, I didn't see the need to make it available to a wider scope.  Though I was thinking of refactoring that so the ignored blocker list is passed in.  If we did that, we could refactor _report_leapp_failure_and_die() to use search_report_file_for_blockers() find all the blockers reported as a result of failure.  As it stands it only prints out information on the blocker to do with a problem with the answer file (which we preemptively fix).  It doesn't filter based in the "inhibitor" flag.

lib/Elevate/Leapp.pm
```diff
@@ -51,13 +51,29 @@ sub install ($self) {
...
     }
 
     return;
 }
 
+sub preupgrade ($self) {
+
+    INFO("Running leapp preupgrade checks");
+
+    $self->cpev->ssystem_hide_output( '/usr/bin/leapp', 'preupgrade' );
```
cPholloway, line 70: Do we need the `--nowarn` flag for CloudLinux here?  It was required for `upgrade`, but I'm not sure about `preupgrade`

script/elevate-cpanel.PL
```diff
@@ -1080,6 +1081,7 @@ sub run_stage_3 ($self) {
     $self->run_once(
         install_leapp => sub {
             $self->leapp->install();
+            $self->leapp->remove_kernel_devel();
```
cPholloway, line 1084: Good call!!! Thanks for separating this out to its own sub

lib/Elevate/Leapp.pm
```diff
@@ -87,6 +103,70 @@ sub upgrade ($self) {
     return;
 }
 
+sub search_report_file_for_blockers ( $self, @ignored_blockers ) {
```
cPholloway, line 106: Out of scope for this case, but we should refactor  `_report_leapp_failure_and_die()` to use this so that we provide info about all the inhibitors to the customer in the event that leapp upgrade fails.

@timmullin - Would you mind filing a follow up case to refactor `_report_leapp_failure_and_die()` to use `search_report_file_for_blockers()` please?

cPholloway (changes requested):

lib/Elevate/Blockers.pm
```diff
@@ -121,7 +122,7 @@ sub _has_blockers ( $self, $check_mode = 0 ) {
     }
 
     $_CHECK_MODE = !!$check_mode;    # running with --check
-    $self->abort_on_first_blocker( !$_CHECK_MODE );
+    $self->abort_on_first_blocker(0);
```
toddr, line 125: Was this testing to remove `!$_CHECK_MODE`?

timmullin: No.  For a while we had been always been calling _has_blockers with check mode enabled.  Check mode was always true and abort_on_first_blocker() always was initialized with 0.  The abort_on_first_blocker() would later be called with true for certain cases like when upcp was running.
I started calling this with check mode as "false" if called via --start.  I needed this so I would know whether to skip the leapp blocker.  But, I did not want to change the behavior with regards to blocking on the first blocker.

lib/Elevate/Leapp.pm
timmullin: I'm not sure about that one.

lib/Elevate/Blockers.pm
```diff
@@ -68,6 +68,7 @@ our @BLOCKERS = qw{
   OVH
   Python
   AbsoluteSymlinks
+  Leapp
```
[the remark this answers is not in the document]
toddr: It would have to come after with a commented push onto the array if you want to assure that.

lib/Elevate/Leapp.pm
cPholloway: Sounds good.  Thank you for confirming this 👍

lib/Elevate/Leapp.pm
timmullin: Filed as RE-217

lib/Elevate/Blockers.pm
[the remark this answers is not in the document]
timmullin: I'll add a comment up at the top where it says "order matters".  I don't see any good place in the module to do the commented push.

cPholloway (approved):

lib/Elevate/Blockers.pm
```diff
@@ -68,6 +69,7 @@ our @BLOCKERS = qw{
...
 };
```
toddr, line 74:
```perl
push @BLOCKERS, 'Leapp'; # This blocker has to run last!
```

timmullin: Okay, I'll add that.

toddr (approved):

cPholloway (approved):

PR #377, opened 2024-02-20
Do not remove alt-pcre802 post elevate

cPholloway:
Case RE-206:  It was found that there are ea-php51 and ea-php52 packages including ea-php51 and ea-php52 that are dependent on alt-pcre802 and alt-pcre802-devel.  As such, we are going to refrain from removing this two packages after elevate for now.  RE-213 was created to follow up with a more thorough fix for this issue as we should refactor this logic so that the leapp packages are separated from the rest of the el7 packages that we should be able to safely remove.

Changelog:  Do not remove 'alt-pcre802' post elevate


annes449 (commented):
Perfect! QA is good to go

PR #378, opened 2024-02-21
Fix PostgreSQL user database check when feature disabled

timmullin:
Case RE-188:

Instead of, rather inefficiently, invoking uapi for each user, get the users with postgres databases from the dbindex.db.json file. This will quickly get that list of users regardless of the status of the postgres feature.

Changelog: Improve the check for PostgreSQL users to be more
  efficient and to not throw errors if the feature is disabled
  for any of the users.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -476,11 +476,13 @@ BEGIN {    # Suppress load of all of these at earliest point.
 
     use cPstrict;
 
-    use Cpanel::OS              ();
-    use Cpanel::Pkgr            ();
-    use Cpanel::Version::Tiny   ();
-    use Cpanel::JSON            ();
-    use Cpanel::SafeRun::Simple ();
+    use Cpanel::OS                         ();
+    use Cpanel::Pkgr                       ();
+    use Cpanel::Version::Tiny              ();
+    use Cpanel::JSON                       ();
+    use Cpanel::SafeRun::Simple            ();
+    use Cpanel::DB::Map::Collection::Index ();
+    use Cpanel::Exception                  ();
 
     # use Elevate::Blockers::Base();
     our @ISA;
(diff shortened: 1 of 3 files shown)


elevate-cpanel
```diff
@@ -477,11 +477,14 @@ BEGIN {    # Suppress load of all of these at earliest point.
...
-    use Cpanel::JSON            ();
-    use Cpanel::SafeRun::Simple ();
+    use Cpanel::OS                         ();
+    use Cpanel::Pkgr                       ();
+    use Cpanel::Version::Tiny              ();
+    use Cpanel::JSON                       ();
+    use Cpanel::SafeRun::Simple            ();
+    use Cpanel::DB::Map::Collection::Index ();
+
+    # use Cpanel::Exception                  ();
```
toddr, line 487: needed?

toddr (changes requested):

elevate-cpanel
timmullin: It is strange that Cpanel::Exception got commented out in the main script.  I included it in lib/Elevate/Blockers/Databases.pm because it invokes Cpanel::Exception::get_string().

cPholloway: This definitely looks like a bug in `main/perlpkg.static`:

```
2880 # if Cpanel::Exception::CORE is included also consider that Cpanel::Exception is in
2881 if ( grep { 'Cpanel::Exception::CORE' } @packages_injected ) {
2882     push @packages_injected, 'Cpanel::Exception';
2883 }
```

Pretty sure that is just `if ( scalar @packages_injected ) { ... }` which is likely not what we want.

Guessing it should be:

```
--- a/maint/perlpkg.static
+++ b/maint/perlpkg.static
@@ -2878,7 +2878,7 @@ foreach my $module ( sort keys %EMBEDDED_NON_CORE_MODULES ) {
 my @packages_injected = map { my $name = $_; $name =~ s{/}{::}g; $name =~ s{\.pm$}{}; $name } @modules_to_include;

 # if Cpanel::Exception::CORE is included also consider that Cpanel::Exception is in
-if ( grep { 'Cpanel::Exception::CORE' } @packages_injected ) {
+if ( grep { $_ eq 'Cpanel::Exception::CORE' } @packages_injected ) {
     push @packages_injected, 'Cpanel::Exception';
 }
```

lib/Elevate/Blockers/Databases.pm
```diff
@@ -89,28 +91,15 @@ sub _blocker_acknowledge_postgresql_datadir ($self) {
...
-        my $dbs_parsed = Cpanel::JSON::Load($dbs_json);
-        if ( !$dbs_parsed->{result}->{status} ) {
-            return if ( $dbs_parsed->{result}->{errors}->[0] // '' ) eq 'This server does not support this functionality.';    # sanity check
-            die( "UAPI Postgresql::list_databases call returned status 0 for user $user:\n" . join( "\n", $dbs_parsed->{result}->{errors}->@* ) );
-        }
-        push @users_with_dbs, $user if scalar $dbs_parsed->{result}->{data}->@*;
+    my $dbindex = eval { Cpanel::DB::Map::Collection::Index->new( { db => 'PGSQL' } ); };
+    if ( my $exception = $@ ) {
+        ERROR( 'Unable to read the database index file:  ' . Cpanel::Exception::get_string($exception) );
+        return ();
```
cPholloway, line 97: We should probably block if we hit this exception.  This would also be an indication that the mysql reinstall is going to go very poorly since the mysql user map is contained in this file too, correct?

cPholloway (changes requested):

lib/Elevate/Blockers/Databases.pm
timmullin: It'd get rebuilt when /ulc/bin/dbindex gets run.

timmullin: Yeah, there should be a blocker for this.  It is an unusual situation; but, it also would mean something is really awry.

elevate-cpanel
toddr: I think that's the same code. Instead, I think you need to remove the injection altogether.

timmullin: Should we also remove it from Blockers/BootKernel.pm?  It also includes Cpanel::Exception, and with the same result (its "use Cpanel::Exception" line is also commented out in elevate-cpanel.)

toddr: This was fixed by c65415117db3c4f151330b35c4115e06fed3c178 followed by a rebase.

cPholloway (approved):

annes449 (approved):
QA is all good.

PR #379, opened 2024-02-21
Add blocker for invalid CloudLinux licenses

cPholloway:
Case RE-173:  CloudLinux's fork of ELevate verifies that the CloudLinux licese is valid and inhibits the elevation process from starting if it is not.  Due to this, we need to block if the CloudLinux license is not valid in order to avoid leaving the system in a broken state.

Additionally, I fixed a bug where the argument would not be numeric in exit when blockers were set to abort on the first blocker, and a bug where the blocker was not added before dying in this same situation which would result in the elevate-blockers file essentially being null.

Changelog: Add blocker for invalid CloudLinux licenses

### DIFF
--- a/lib/Elevate/Blockers/CloudLinux.pm
+++ b/lib/Elevate/Blockers/CloudLinux.pm
@@ -1,0 +1,47 @@
+package Elevate::Blockers::CloudLinux;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Blockers::CloudLinux
+
+Blocker to check that the CloudLinux licese is valid for CloudLinux upgrades only.
+
+=cut
+
+use cPstrict;
+
+use parent qw{Elevate::Blockers::Base};
+
+use Elevate::OS ();
+
+use Log::Log4perl qw(:easy);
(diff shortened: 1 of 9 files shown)


lib/Elevate/Blockers/CloudLinux.pm
```diff
@@ -0,0 +1,47 @@
...
+sub check ($self) {
+    return $self->_check_cloudlinux_license();
+}
+
+sub _check_cloudlinux_license ($self) {
+    return 0 unless Elevate::OS::should_check_cloudlinux_license();
+
+    my $out = $self->ssystem_capture_output( CLDETECT, '--check-license' );
+
+    if ( $self->ssystem(RHN_CHECK) != 0 || $out->{status} != 0 || grep { $_ !~ m/^ok/i } @{ $out->{stdout} } ) {
```
toddr, line 33: if you do `return 0 unless`, then you don't need a big block.

toddr (approved):

annes449 (approved):
QA is all good

PR #381, opened 2024-02-27
Re 89 fixups

cPholloway:

### DIFF
--- a/lib/Elevate/Blockers/Base.pm
+++ b/lib/Elevate/Blockers/Base.pm
@@ -93,10 +93,20 @@ sub has_blocker ( $self, $msg, %others ) {
 
     my $analytics_data;
     if ( $others{info} ) {
-        my $info = $others{info};
-        die "Invalid data analytics given to blocker.  'info' must be a hash reference.\n" unless ref $info eq 'HASH';
-        $analytics_data = Cpanel::JSON::canonical_dump( [$info] );
-        delete $others{info};
+        my $info = delete $others{info};
+
+        if ( ref $info eq 'HASH' ) {
+            $analytics_data = Cpanel::JSON::canonical_dump( [$info] );
+        }
+        else {
+
+            # only die if this is a developement version since we do not analytics data to potentially affect
+            # the ability to elevate in production
+            my ( $latest_version, $self_version ) = ( $self->cpev->script->latest_version(), cpev::VERSION() );
+            if ( $self_version > $latest_version ) {
+                die "Invalid data analytics given to blocker.  'info' must be a hash reference.\n";
+            }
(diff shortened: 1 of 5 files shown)


lib/Elevate/Blockers/Base.pm
```diff
@@ -90,7 +91,15 @@ sub has_blocker ( $self, $msg, %others ) {
         $caller_id ||= ref $self;
     }
 
-    my $blocker = cpev::Blocker->new( id => $caller_id, msg => $msg, %others );
+    my $analytics_data;
+    if ( $others{info} ) {
+        my $info = $others{info};
```
timmullin, line 96: Not a blocker or anything; but, I believe you could do:
my $info = delete $others{info};
Then you wouldn't need line 99.

lib/Elevate/Blockers/Repositories.pm
```diff
@@ -63,24 +64,19 @@ sub _blocker_invalid_yum_repos ($self) {
 
         for my $unsupported_repo ( @{ $self->{_yum_repos_unsupported_with_packages} } ) {
             my $blocker_id = ref($self) . '::' . $unsupported_repo->{'name'};
-            $self->has_blocker( $unsupported_repo->{'json_report'}, 'blocker_id' => $blocker_id, 'quiet' => 1 );
+            $self->has_blocker(
+                $msg,
```
timmullin, line 68: It appears that if neither `INVALID_SYNTAX` or `USE_RPMS_FROM_UNVETTED_REPO` are set, then the `$msg` will be empty.  There probably ought to be a default message.

cPholloway: This code is unreachable unless either `INVALID_SYNTAX` or `USE_RPMS_FROM_UNVETTED_REPO` are set.

` 63         return 0 unless _yum_status_hr_contains_blocker($status_hr);`

^ This will return prior to reaching this code if neither are set.

lib/Elevate/Blockers/Base.pm
cPholloway: Good call.  I'll make this change if/when I fix something else up.  Not going to bother if this is the only requested change.

lib/Elevate/Blockers/Repositories.pm
timmullin: Okay, cool.

timmullin (approved):

toddr (approved):

toddr (changes requested):
need to fix the assert and change back to include package names.

annes449 (approved):
QA is all good.

PR #382, opened 2024-02-29
Do not copy the stage file to the success file until we notify success

cPholloway:
Case RE-228: The success file was reporting that elevate was still running.  This occurred because we were copying the success file into place in stage 5 before elevate had been marked as being succesful. This change makes it so that we do not copy the stage file to the success file until after we have marked in the stage file that we succeeded and notified the customer that the process succeeded.

Changelog: Do not copy the stage file to the success file until we
 notify the user that the process succeeded


toddr (approved):

annes449 (approved):
QA is all good.

PR #388, opened 2024-03-07
Block if sectigo is the provider for autossl domains

timmullin:
Case RE-54:

When CentOS 7 is elevated to AlmaLinux 8, the "nobody" user will retain its uid/gid of 99 where as on AlmaLinux 8 the expected uid/gid for "nobody" is 65534. This will cause issues with the Sectigo AutoSSL provider.

Changelog: Block if Sectigo is the AutoSSL provider.

### DIFF
--- a/lib/Elevate/Blockers/AutoSSL.pm
+++ b/lib/Elevate/Blockers/AutoSSL.pm
@@ -1,0 +1,48 @@
+package Elevate::Blockers::AutoSSL;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Blockers::AutoSSL
+
+Blocker to check if Sectigo is the AutoSSL provider.
+
+=cut
+
+use cPstrict;
+
+use Cpanel::SSL::Auto ();
+
+use parent qw{Elevate::Blockers::Base};
+
+sub check ($self) {
+
+    return $self->_check_autossl_provider();
+}
+
+sub _check_autossl_provider ($self) {
+
+    my @providers = Cpanel::SSL::Auto::get_all_provider_info();
+
+    foreach my $provider (@providers) {
+        next unless ( ref $provider eq 'HASH' && $provider->{enabled} );
+
+        if ( defined $provider->{display_name}
+            && $provider->{display_name} =~ /sectigo/i ) {
+            return $self->has_blocker( <<~"EOS");
+            Elevating with the $provider->{display_name} provider in place is no longer supported.
+            To switch to the Let's Encrypt™ provider, review their terms of service here:
+            https://letsencrypt.org/documents/LE-SA-v1.3-September-21-2022.pdf
+            Then, if you accept, execute this command:
+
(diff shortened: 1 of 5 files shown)


toddr (approved):

lib/Elevate/Blockers/AutoSSL.pm
```diff
@@ -0,0 +1,43 @@
...
+    my @providers = Cpanel::SSL::Auto::get_all_provider_info();
+
+    foreach my $provider (@providers) {
+        next unless ( ref $provider eq 'HASH' && $provider->{enabled} );
+
+        if ( defined $provider->{display_name}
+            && $provider->{display_name} =~ /sectigo/i ) {
+            return $self->has_blocker( <<~"EOS");
+            Unsupported AutoSSL provider: $provider->{display_name}
+            Move all accounts to a different AutoSSL provider before running elevate.
```
cPholloway, line 35: Is there an API call that we can provide the customer to make it easy for them to do this / would `set_autossl_provider` work for this purpose?

cPholloway (approved):

lib/Elevate/Blockers/AutoSSL.pm
timmullin: ```set_autossl_provider``` should do it.  The might also need to invoke ```start_autossl_check_for_all_users```; but, I think that happens automatically on a regular basis.

annes449 (approved):
QA is all good.

lib/Elevate/Blockers/AutoSSL.pm
```diff
@@ -0,0 +1,46 @@
...
+        next unless ( ref $provider eq 'HASH' && $provider->{enabled} );
+
+        if ( defined $provider->{display_name}
+            && $provider->{display_name} =~ /sectigo/i ) {
+            return $self->has_blocker( <<~"EOS");
+            Elevating with the $provider->{display_name} provider in place is no longer supported.
+            To switch to the Let's Encrypt™ provider, review their terms of service here:
+            https://letsencrypt.org/documents/LE-SA-v1.3-September-21-2022.pdf
+            Then, if you accept, execute this command:
+            'whmapi1 set_autossl_provider provider='LetsEncrypt' x_terms_of_service_accepted=https%3A%2F%2Fletsencrypt.org%2Fdocuments%2FLE-SA-v1.3-September-21-2022.pdf'
```
cPholloway, line 38: Can we use the absolute path for whmpi1 here:  `/usr/local/cpanel/bin/whmapi1`

Also, can we get rid of the single quotes surrounding the command and instead put a blank line to either side of the command to make it easier to distinguish and copy/paste

cPholloway (changes requested):

cPholloway (approved):

annes449 (approved):
I think this is great. QA is all good.

toddr:
I'm declining this for now.

PR #391, opened 2024-03-12
Block if grub not installed or config file missing

timmullin:
Case RE-83:

Changelog: Block if the grub2-pc package is not installed
  or if the grub.cfg file is missing.

### DIFF
--- a/lib/Elevate/Blockers/Grub2.pm
+++ b/lib/Elevate/Blockers/Grub2.pm
@@ -12,6 +12,8 @@ Blocker to check compatibility with Grub2 configuration.
 
 use cPstrict;
 
+use Cpanel::Pkgr ();
+
 use Elevate::Constants ();
 
 use parent qw{Elevate::Blockers::Base};
@@ -36,6 +38,8 @@ sub check ($self) {
     my $ok = 1;
     $ok = 0 unless $self->_blocker_grub2_workaround;
     $ok = 0 unless $self->_blocker_blscfg;
+    $ok = 0 unless $self->_blocker_grub_not_installed;
+    $ok = 0 unless $self->_blocker_grub_config_missing;
     return $ok;
 }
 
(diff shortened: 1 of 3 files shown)


lib/Elevate/Blockers/Grub2.pm
```diff
@@ -99,6 +103,28 @@ sub _blocker_blscfg ($self) {
     return 0;
 }
 
+sub _blocker_grub_not_installed ($self) {
+
+    return 0 if Cpanel::Pkgr::is_installed('grub2-pc');
+
+    return $self->has_blocker( <<~EOS );
+    The grub2-pc package is not installed. The GRUB2 boot loader is
+    required to upgrade via leapp.
```
cPholloway, line 112: Let's advise them to consider reaching out to support if they hit this blocker.  There is already some wording to do this earlier in this file via `_blocker_grub2_workaround()`

lib/Elevate/Blockers/Grub2.pm
```diff
@@ -99,6 +103,28 @@ sub _blocker_blscfg ($self) {
...
+    EOS
+}
+
+sub _blocker_grub_config_missing ($self) {
+
+    if ( !-e '/boot/grub2/grub.cfg' && !-e '/boot/grub/grub.cfg' ) {
+
+        return $self->has_blocker( <<~EOS );
+        The GRUB2 config file is missing.
```
cPholloway, line 121: Let's advise them to consider reaching out to support if they hit this blocker.  There is already some wording to do this earlier in this file via `_blocker_grub2_workaround()`

lib/Elevate/Blockers/Grub2.pm
```diff
@@ -99,6 +103,28 @@ sub _blocker_blscfg ($self) {
...
+
+    return $self->has_blocker( <<~EOS );
+    The grub2-pc package is not installed. The GRUB2 boot loader is
+    required to upgrade via leapp.
+    EOS
+}
+
+sub _blocker_grub_config_missing ($self) {
+
+    if ( !-e '/boot/grub2/grub.cfg' && !-e '/boot/grub/grub.cfg' ) {
```
cPholloway, line 118: We should use `-f` instead of `-e` here.  Possibly `-s` too to make sure it is not a touch file.

cPholloway (changes requested):

cPholloway (approved):

toddr (approved):

annes449 (commented):
QA looks good.

rch77:
Now it blocks if there is /boot/efi/EFI/centos/grub.cfg only. I think UEFI systems use /boot/efi/EFI/centos/grub.cfg.
I have added symlink from /boot/grub2/ to ../efi/EFI/centos/grub.cfg but I'm not sure if it'll break something else during the process.
OS: Centos 7.9

PR #392, opened 2024-03-12
Remove blocker to check if the system is up to date

cPholloway:
Case RE-172:  In stage 2, we execute 'yum -y update' and die if the system fails to update.  Given that we are already ensuring that the system is up to date in stage 2, there is no need to block if the system is up to date as it redundant to a task that we are performing in stage
2.  As such, this blocker is not necessary.

Changelog: Remove blocker to check if the system is up to date


timmullin (approved):

annes449 (approved):
QA is all good.

PR #393, opened 2024-03-12
Ensure upcp and backups cannot run at the same time as ELevate.

cPanelJossieBrandt:
Case RE-138: This commit makes it so scripts/upcp and bin/backup are removed early in stage 2 so cron or another entity cannot run them during the ELevate process.

Changelog: Ensure upcp and backups cannot run at the same time
 as ELevate.

### DIFF
--- a/script/elevate-cpanel.PL
+++ b/script/elevate-cpanel.PL
@@ -1038,6 +1038,11 @@ sub run_stage_2 ($self) {
 
     $self->elevation_startup_marker($sum);
 
+    # Remove scripts/upcp and bin/backup to make sure they don't
+    # end up running while ELevate is running.
+    unlink('/usr/local/cpanel/scripts/upcp');
+    unlink('/usr/local/cpanel/bin/backup');
+
     $self->ssystem(qw{/usr/bin/yum clean all});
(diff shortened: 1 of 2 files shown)


script/elevate-cpanel.PL
```diff
@@ -1038,6 +1038,11 @@ sub run_stage_2 ($self) {
 
     $self->elevation_startup_marker($sum);
 
+    # Remove scripts/upcp and bin/backup to make sure they don't
+    # end up running while ELevate is running.
+    $self->ssystem(qw{/usr/bin/rm -f /usr/local/cpanel/scripts/upcp});
+    $self->ssystem(qw{/usr/bin/rm -f /usr/local/cpanel/bin/backup});
```
timmullin, line 1044: We ought to handle the case where these fail.  Perhaps ```ssystem_and_die()``` would suffice.

toddr: Could we just do `unlink('/LC/bin/backup')` here please? There are far fewer things that can go wrong with that.

toddr (changes requested):

toddr (approved):

toddr:
This is replaced by #399

PR #399, opened 2024-03-25
Ensure upcp and backups cannot run at the same time as ELevate.

toddr:
Case RE-138: This commit makes it so scripts/upcp and bin/backup are removed early in stage 2 so cron or another entity cannot run them during the ELevate process.

Changelog: Ensure upcp and backups cannot run at the same time
 as ELevate.


cPholloway (approved):

timmullin (approved):

annes449 (approved):
QA is all good.

PR #400, opened 2024-03-25
Convert systems using Sectigo to use Let's Encrypt

timmullin:
Case RE-54:

When CentOS 7 is elevated to AlmaLinux 8, the "nobody" user will retain its uid/gid of 99 where as on AlmaLinux 8 the expected uid/gid for "nobody" is 65534. This will cause issues with the Sectigo AutoSSL provider.
If a system is setup to use Sectigo as the AutoSSL provider, emit a warning that the system will be switched to Lets Encrypt. Run the appropriate autofixer for the conversion during the upgrade

Changelog: Change to Let's Encrypt if Sectigo is the AutoSSL provider.

### DIFF
--- a/script/elevate-cpanel.PL
+++ b/script/elevate-cpanel.PL
@@ -278,6 +278,7 @@ use Elevate::Blockers::Repositories     ();    # using a constant
 use Elevate::Blockers::SSH              ();
 use Elevate::Blockers::WHM              ();
 use Elevate::Blockers::Leapp            ();
+use Elevate::Blockers::AutoSSL          ();
 
 # - fatpack Components
 use Elevate::Components::Base               ();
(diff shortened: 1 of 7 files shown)


script/elevate-cpanel.PL
```diff
@@ -1052,6 +1054,7 @@ sub run_stage_2 ($self) {
 
     Elevate::Motd->setup();
 
+    $self->run_component_once( 'AutoSSL'    => 'pre_leapp' );
```
cPholloway, line 1057: The 'SSH' component should still get executed before this.  Can we move this down one line please?

cPholloway:
Can we add a test for the new component as well please?

cPholloway (changes requested):

script/elevate-cpanel.PL
timmullin: Okay, I'll move it to run after the SSH component.

timmullin:
> Can we add a test for the new component as well please?

I can go ahead and add one.

cPholloway (approved):

toddr (approved):

annes449 (approved):
All good with QA

PR #401, opened 2024-03-26
Validate script can execute start option before parsing upgrade_to

cPholloway:
Case RE-265: The start option had an order of operations issue where it would attempt to parse the upgrade_to option before validating that it could start / was not already running.  Since parsing the upgrade_to option validates that the distro is appropriate, it would clear the Elevate::OS cache which could result in the script dying once leapp finished elevating.  This change ensures that the validation that the script can start occurs before parsing the upgrade_to option.

Changelog: Validate script can execute start option before parsing
 upgrade_to option

### DIFF
--- a/lib/Elevate/OS.pm
+++ b/lib/Elevate/OS.pm
@@ -57,6 +57,8 @@ sub instance {
         die "This script is only designed to upgrade the following OSs:\n\n$supported_distros\n";
     }
 
+    Elevate::OS::_set_cache();
+
     return $OS;
 }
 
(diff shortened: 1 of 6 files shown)


timmullin (approved):

lib/Elevate/OS.pm
```diff
@@ -57,6 +57,8 @@ sub instance {
         die "This script is only designed to upgrade the following OSs:\n\n$supported_distros\n";
     }
 
+    Elevate::OS::_set_cache();
```
toddr, line 60: It feels like you're risking a loop here. Why not just run `_set_cache($OS)`

toddr (approved):

annes449 (approved):
QA is all good

PR #406, opened 2024-03-29
Update warning string in last chance message

cPholloway:
Case RE-274: The warning string in the last chance message was hard coded to CentOS 7.  Since we now support multiple OSs, this string needs to be updated to be a variable showing the proper OS that is being updated.

Changelog:  Update string in last chance message to inform the user of
 the correct OS that is being updated


toddr (approved):

annes449 (approved):
Qa is all good!

PR #410, opened 2024-04-06
Handle auto-upgrade of out-of-date MySQL installs

timmullin:
Case RE-171:

Instead of simply having a blocker for old versions of MySQL/MariaDB, give the option to auto-upgrade
during the elevate process. Also, give them the option to specify which version to which they prefer to upgrade. We can only do this when MySQL/MariaDB is provided by cPanel, not CloudLinux.

Changelog: Provide auto-upgrade mechanism for out-of-date
  MySQL/MariaDB versions (installed via cPanel, not CloudLinux).

### DIFF
--- a/lib/Elevate/Blockers/Base.pm
+++ b/lib/Elevate/Blockers/Base.pm
@@ -17,7 +17,6 @@ use Cpanel::JSON ();
 
 use Simple::Accessor qw(
   blockers
-  cpconf
 );
 
 use Log::Log4perl qw(:easy);
@@ -56,10 +55,6 @@ BEGIN {
             return $sub->( $cpev, @args );
         }
     }
-}
-
-sub _build_cpconf ($self) {
-    return Cpanel::Config::LoadCpConf::loadcpconf() // {};
 }
 
 =head2 $self->is_check_mode( @args )
--- a/lib/Elevate/Blockers/Databases.pm
+++ b/lib/Elevate/Blockers/Databases.pm
@@ -22,6 +22,7 @@ use Cpanel::JSON                       ();
 use Cpanel::SafeRun::Simple            ();
 use Cpanel::DB::Map::Collection::Index ();
 use Cpanel::Exception                  ();
+use Cpanel::MysqlUtils::MyCnf::Basic   ();
 
 use parent qw{Elevate::Blockers::Base};
 
@@ -33,6 +34,7 @@ sub check ($self) {
     my $ok = 1;
     $self->_warning_if_postgresql_installed;
     $ok = 0 unless $self->_blocker_acknowledge_postgresql_datadir;
+    $ok = 0 unless $self->_blocker_remote_mysql;
     $ok = 0 unless $self->_blocker_old_mysql;
     $ok = 0 unless $self->_blocker_mysql_upgrade_in_progress;
     $self->_warning_mysql_not_enabled();
@@ -105,6 +107,23 @@ sub _has_mapped_postgresql_dbs ($self) {
     return ( keys %user_hash );
 }
 
+sub _blocker_remote_mysql ($self) {
+
+    my $pretty_distro_name = $self->upgrade_to_pretty_name();
+
+    # If we are setup to use remote MySQL, then attempting an upgrade will fail
+    # TODO: Temporarily disable remote MySQL to allow the database upgrade
+    if ( Cpanel::MysqlUtils::MyCnf::Basic::is_remote_mysql() ) {
+        return $self->has_blocker( <<~"EOS" );
+        The system is currently setup to use a remote database server.
+        We cannot elevate the system to $pretty_distro_name
+        unless the system is configured to use the local database server.
+        EOS
+    }
+
+    return 0;
+}
+
 sub _blocker_old_mysql ($self) {
 
     my $mysql_is_provided_by_cloudlinux = Elevate::Database::is_database_provided_by_cloudlinux(0);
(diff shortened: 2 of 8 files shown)


lib/Elevate/Components/Database.pm
cPholloway, line 1: FIX:

I see why this needs to be separate from `Elevate::Components::MySQL` since it needs to run much earlier than the `Elevate::Components::MySQL`.  However, to avoid confusion between this and `Elevate::Component::MySQL`, we should change the name of this to be more specific to what it is doing.  Maybe something like:

`Elevate::Components::DatabaseUpgrade`

t/Elevate-Database.t
```diff
@@ -159,4 +160,103 @@ $mock_stagefile->redefine(
...
+
+    foreach my $fail_version (@should_fail) {
+        ok( !Elevate::Database::is_database_version_supported($fail_version), "$fail_version is NOT supported" );
+    }
+}
+
+{
+    note 'Test validate_mysql_upgrade_version() behavior';
+
+    my $current_version  = '10.5';
```
cPholloway, line 218: FIX:

Ensure that combinations such as 5.6->5.7 or 10.0->10.2 are disallowed.  This is technically okay if they were staying on C7 but it will blow up on A8 so we need to be very cautious about blocking something like that as it will result in a broken system after elevate completes.

lib/Elevate/Database.pm
```diff
@@ -75,4 +87,64 @@ sub get_db_info_if_provided_by_cloudlinux ( $use_cache = 1 ) {
...
+    require Whostmgr::Mysql::Upgrade;
+
+    return Whostmgr::Mysql::Upgrade::get_latest_available_version( version => get_local_database_version() );
+}
+
+sub get_database_type_name_from_version ($version) {
+    return Cpanel::MariaDB::version_is_mariadb($version) ? 'MariaDB' : 'MySQL';
+}
+
+sub validate_mysql_upgrade_version ($upgrade_version) {
```
cPholloway, line 125: FIX:

This would still allow something like upgrading MySQL 5.6 to MySQL 5.7 which would cause an elevate failure.  Additionally, after the initial check from lines 130-145 completes, we should run the version through `Elevate::Database::is_database_version_supported` to validate that the mysql target is an eligible upgrade target for elevate too.  In other words, 

1.  Confirm that cPanel can/will do the upgrade
2. Confirm that the version is valid for elevate

cPholloway (changes requested):

lib/Elevate/Blockers/Databases.pm
```diff
@@ -141,67 +142,77 @@ sub _blocker_old_cloudlinux_mysql ($self) {
...
-            You are using MySQL $mysql_version server.
-            This version is not available for $pretty_distro_name.
-            You first need to update your MySQL server to 8.0 or later.
+    # If we are running a local version of MySQL/MariaDB that will be
+    # supported by the new OS version, we leave it as it is.
+    if ( Elevate::Database::is_database_version_supported($mysql_version) ) {
 
-            You can update to version 8.0 using the following command:
+        # store the MySQL version we started from
+        Elevate::StageFile::update_stage_file( { 'mysql-version' => $mysql_version } );
```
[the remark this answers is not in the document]
timmullin: This line is there to preserve the original behavior.  In the main branch, if we do not hit a blocker due to the MySQL version, then we update the stage file with the current version.

lib/Elevate/Blockers/Databases.pm
```diff
@@ -141,67 +142,77 @@ sub _blocker_old_cloudlinux_mysql ($self) {
...
+    WARN( <<~"EOS" );
+    Prior to elevating this system to $pretty_distro_name,
+    we will automatically upgrade your installation of $database_type_name
+    to $upgrade_dbtype_name $upgrade_version.
+    If you would prefer to upgrade $database_type_name to a different version,
+    you can rerun this script with --mysql=<upgrade version>
+
+    EOS
+
+    if (
```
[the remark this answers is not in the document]
timmullin: Yeah, that makes sense.  I didn't think of that.

lib/Elevate/Blockers/Databases.pm
```diff
@@ -141,67 +142,77 @@ sub _blocker_old_cloudlinux_mysql ($self) {
...
+        !IO::Prompt::prompt(
+            '-one_char',
+            '-yes_no',
+            '-tty',
+            -default => 'n',
+            "Do you consent to upgrading to $upgrade_dbtype_name $upgrade_version [y/N]: ",
+        )
+    ) {
+        return $self->has_blocker( <<~"EOS" );
+        The system cannot be elevated to $pretty_distro_name until $database_type_name has been upgraded.
```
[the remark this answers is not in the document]
timmullin: I'll go ahead and make that change; I think your way of wording it would be more clear to the user.

lib/Elevate/Components/Database.pm
```diff
@@ -0,0 +1,68 @@
+package Elevate::Components::Database;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::Database
+
+Change Database provider from Sectigo to Lets Encrypt
```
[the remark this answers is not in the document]
timmullin: Copy Pasta.  My bad.

lib/Elevate/Blockers/Databases.pm
```diff
@@ -141,67 +142,77 @@ sub _blocker_old_cloudlinux_mysql ($self) {
...
+    you can rerun this script with --mysql=<upgrade version>
+
+    EOS
+
+    if (
+        !IO::Prompt::prompt(
+            '-one_char',
+            '-yes_no',
+            '-tty',
+            -default => 'n',
```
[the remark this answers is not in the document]
timmullin: Okay.

lib/Elevate/Usage.pm
```diff
@@ -24,6 +26,7 @@ sub _OPTIONS {
       no-leapp
       non-interactive
       leappbeta
+      mysql=s
```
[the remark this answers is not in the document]
timmullin: I am about to eliminate the "--mysql" parameter from this pull request.  It was decided that giving multiple upgrade options will complicate testing.

lib/Elevate/Components/Database.pm
timmullin: Makes sense; I'll do that.

t/Elevate-Database.t
timmullin: This is about to be dropped.

lib/Elevate/Blockers/Databases.pm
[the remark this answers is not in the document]
cPholloway: ah, that makes sense.  Consider this resolved.

lib/Elevate/Blockers/Base.pm
```diff
@@ -58,10 +58,6 @@ BEGIN {
     }
 }
 
-sub _build_cpconf ($self) {
```
cPholloway, line 61: FIX:

Need to remove it from the `Simple::Accessor` array too

lib/Elevate/Database.pm
```diff
@@ -15,10 +15,22 @@ use cPstrict;
...
-use Cpanel::Pkgr ();
+use Cpanel::MysqlUtils::Version  ();
+use Cpanel::MysqlUtils::Versions ();
+use Cpanel::Pkgr                 ();
+
+use Log::Log4perl qw(:easy);
 
 use constant MYSQL_BIN => '/usr/sbin/mysqld';
 
+use constant SUPPORTED_MYSQL_VERSIONS => qw{
```
cPholloway, line 26: QUESTION:

Should we update this name to be `SUPPORTED_CPANEL_MYSQL_VERSIONS` since CL does its own thing?

cPholloway (changes requested):

lib/Elevate/Database.pm
timmullin: Good idea.  I'll do that.

cPholloway (approved):

annes449 (approved):
Qa is all good.

cPholloway (approved):

lib/Elevate/Blockers/Databases.pm
```diff
@@ -105,6 +107,23 @@ sub _has_mapped_postgresql_dbs ($self) {
     return ( keys %user_hash );
 }
 
+sub _blocker_remote_mysql ($self) {
+
+    my $pretty_distro_name = $self->upgrade_to_pretty_name();
```
toddr, line 112: you could technically move this into the conditional. :)

lib/Elevate/Database.pm
```diff
@@ -15,10 +15,22 @@ use cPstrict;
...
+use Cpanel::MysqlUtils::Versions ();
+use Cpanel::Pkgr                 ();
+
+use Log::Log4perl qw(:easy);
 
 use constant MYSQL_BIN => '/usr/sbin/mysqld';
 
+use constant SUPPORTED_CPANEL_MYSQL_VERSIONS => qw{
+  8.0
+  10.3
```
toddr, line 28: I could swear 10.3 was NOT supported on 8? :\

toddr (commented):
comments

PR #411, opened 2024-04-10
Convert CCS blocker to component

cPholloway:
Case RE-78: This change converts the CCS blocker to a component.

    * Remove the CCS blocker
    * Export the CCS data to a secure directory prior to starting leapp
    * Remove the CCS package prior to starting leapp
    * Remove some cruft that the CCS package leaves behind prior to leapp
    * Install the CCS package after leapp completes
    * Import the CCS data after leapp completes

    Changelog: Convert CCS blocker to component

### DIFF
--- a/lib/Elevate/Components/CCS.pm
+++ b/lib/Elevate/Components/CCS.pm
@@ -1,0 +1,461 @@
+package Elevate::Components::CCS;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::CCS
+
+pre_leapp: Export CCS data to a root owned backup directory and remove CCS
+           package
+
+post_leapp: Install CCS package and import the backups taken during pre_leapp
+
+=cut
+
+use cPstrict;
+
+use Try::Tiny;
+
+use File::Path ();
+use File::Copy ();
+
+use Cpanel::Autodie       ();
+use Cpanel::Config::Users ();
+use Cpanel::JSON          ();
+use Cpanel::Pkgr          ();
+
+use Elevate::Notify    ();
+use Elevate::StageFile ();
+
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Components::Base};
+
+use constant CCS_PACKAGE        => 'cpanel-ccs-calendarserver';
+use constant ZPUSH_PACKAGE      => 'cpanel-z-push';
+use constant EXPORT_DIR         => '/var/cpanel/elevate_ccs_export';
+use constant CCS_RESTART_SCRIPT => '/usr/local/cpanel/scripts/restartsrv_cpanel_ccs';
+use constant TASK_QUEUE_SCRIPT  => '/usr/local/cpanel/bin/servers_queue';
+
+use constant DUMP_TYPES => (
+    calendars => 'ics',
+    contacts  => 'vcard',
+);
+
+sub pre_leapp ($self) {
+    my $ccs_installed = Cpanel::Pkgr::is_installed(CCS_PACKAGE);
+    Elevate::StageFile::update_stage_file( { ccs_installed => $ccs_installed } );
+    return unless $ccs_installed;
+
+    $self->_load_ccs_modules();
+
+    $self->run_once('export_ccs_data');
+    $self->remove_ccs_and_dependencies();
+
+    # Removing the PKG will leave this directory in place
+    # This results in PostGreSQL/CCS failing to start after leapp completes
+    $self->clean_up_pkg_cruft();
+
+    return;
+}
+
+sub clean_up_pkg_cruft ($self) {
+    $self->move_pgsql_directory();
+    $self->remove_cpanel_ccs_home_directory();
+    return;
+}
+
+=head1 remove_cpanel_ccs_home_directory
+
+Removing the package removes the `cpanel-ccs` user, but leaves behind
+the data in the home directory.  This results in the service failing
+to start after the elevation has completed
+
+=cut
+
+sub remove_cpanel_ccs_home_directory ($self) {
+    File::Path::remove_tree('/opt/cpanel-ccs') if -d '/opt/cpanel-ccs';
+    return;
+}
+
+=head1 move_pgsql_directory
+
+Removing the PKG will leave this directory in place
+This results in PostGreSQL/CCS failing to start after leapp completes
+
+=cut
+
+sub move_pgsql_directory ($self) {
+    my $pg_dir        = '/var/lib/pgsql';
+    my $pg_backup_dir = '/var/lib/pgsql_pre_elevate';
+
+    # Remove the backup path if it exists as a directory
+    File::Path::remove_tree($pg_backup_dir) if -e $pg_backup_dir && -d $pg_backup_dir;
+
+    # If we were unable to remove the backup path above, then change it to something that
+    # should be unique
+    $pg_backup_dir .= '_' . time() . '_' . $$ if -e $pg_backup_dir;
+
+    # Make sure the path that should be unique does not exist
+    File::Path::remove_tree($pg_backup_dir) if -e $pg_backup_dir && -d $pg_backup_dir;
+
+    # Give it up if we still do not have a candidate to back the data up to
+    if ( -e $pg_backup_dir ) {
+        die <<~"EOS";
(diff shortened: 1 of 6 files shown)


lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,388 @@
...
+
+=head1 move_pgsql_directory
+
+Removing the PKG will leave this directory in place
+This results in PostGreSQL/CCS failing to start after leapp completes
+
+=cut
+
+sub move_pgsql_directory ($self) {
+    File::Copy::move( '/var/lib/pgsql', '/var/lib/pgsql_pre_elevate' ) if -d '/var/lib/pgsql';
```
timmullin, line 89: What if the `/var/lib/pgsql_pre_elevate` directory already exists?

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,388 @@
...
+TL;DR: We will want to monitor CCS for bug fixes/changes as we may need to
+update this code to resolve issues that get resolved there
+
+=cut
+
+sub export_ccs_data ($self) {
+    my $export_dir = EXPORT_DIR();
+    INFO("Exporting CCS data to '$export_dir'.  A backup of this data will be left in place after elevate completes.");
+    $self->_ensure_export_directory();
+
```
timmullin, line 135: I'm wondering long this will take on a system with a large number (say a couple thousand) accounts.  Perhaps we should have another INFO line at the end of the function when it is complete.  Also, maybe an INFO statement every 10-100 accounts or so; something like INFO "Exported 200/3500 accounts".  That way, they won't think elevate has hung during this step.

lib/Elevate/Components/CCS.pm
cPholloway: Good call, I'll go ahead and make sure it does not exist prior to moving the datadir out of the way

lib/Elevate/Components/CCS.pm
cPholloway: Good call.  I'll go ahead and make this change

timmullin (approved):

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,424 @@
...
+    return $self->{$user}{export_path};
+}
+
+sub _get_ccs_info_for_user ( $self, $user ) {
+    $self->{$user}{info} ||= Cpanel::CCS::Userdata::get_cpanel_account_users_uuids($user);
+    return $self->{$user}{info};
+}
+
+sub _ensure_export_directory ($self) {
+    File::Path::make_path(EXPORT_DIR);
```
toddr, line 303: We need to make sure the perms on this path are 700.

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,424 @@
...
+##### post_leapp code below this ###
+####################################
+
+sub post_leapp ($self) {
+    return unless Elevate::StageFile::read_stage_file('ccs_installed');
+
+    $self->dnf->install(CCS_PACKAGE);
+    $self->_ensure_ccs_service_is_up();
+    $self->_clear_task_queue();
+    $self->run_once('import_ccs_data');
```
toddr, line 317: What if the CCS service isn't up. don't we need to fail and abort the script?

toddr: Oh, I see, we're trying to gracefully fail and move on?

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,424 @@
...
+sub remove_ccs_and_dependencies ($self) {
+
+    # There are other dependencies but these are the 3 that we are concerned with
+    my @ccs_depedencies = qw{
+      postgresql
+      postgresql-devel
+      postgresql-server
+    };
+
+    $self->yum->remove( CCS_PACKAGE(), @ccs_depedencies );
```
toddr, line 132: I'm ok with this but I am worried what we'll ALSO remove that we weren't expecting to.

toddr (changes requested):

cPholloway:
^ this is just the rebase on main

lib/Elevate/Components/CCS.pm
cPholloway: done

lib/Elevate/Components/CCS.pm
cPholloway: Correct, it should fail gracefully, and give a final notification of any users that failed at the end of the log.  It will also display the backup directory and recommend contacting support if they need further assistance importing the data at that point.

I chose to do this instead of failing as I don't think that failing to import the CCS should be a fatal error for the script since the data can still be recovered.  I think this is YAGNI for now, but we could expose an option for elevated servers to import the data if it becomes enough of a problem.

lib/Elevate/Components/CCS.pm
cPholloway: We do block on customer owned PG DBs.  The only way we get this far is if there is no customer owned PG data.

troglodyne:
So, my only overall commentary about this is mostly about ways this could have potentially been made faster/simpler. Basically this is just running the pkgacct processes I already had code for and then the restorepkg bits after restore, which should work, but is a bit "inefficient" in our context. Testing on a server with an obnoxious amount of users and data and a server with very few users/data is as such advisable from a performance standpoint. I'm not actually sure whether the amount of inefficiency here will actually matter.

That said, doing the following would have been simpler/more efficient execution-wise (but taken more time to code potentially?):
* Backup ccs_persistence.json from /var/cpanel/ccs/
* Run pg_dump specifying the version we intend to move to. The complication here is having the newest version of pg_dump which understands all these older versions installed.
* Uninstall.
* Elevate.
* Reinstall.
* Flush task queue.
* Stop CCS.
* use `pg_restore` with the `--clean` opt to restore the DB dump onto our cluster.
* Overwrite ccs_persistence.json
* Re-run `/opt/cpanel-ccs/bin/rebuildccsconfig` to reimpose the old UUIDs.
* Start CCS.

troglodyne:
If we wind up needing to do pg_dump for customer DB migration, then we probably *should* revisit this component to share code for that in the future.

troglodyne (approved):

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,424 @@
...
+                $txt =~ tr/'//d;
+                print $dh $txt;
+            }
+        }
+    }
+
+    return;
+}
+
+sub _dump_delegation_data_for_user ( $self, $user ) {
```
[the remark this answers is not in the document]
cPholloway: just lack of overall knowledge of how CCS works.  Most of this is just me porting over code from what pkgacct runs so I didn't realize I could just backup the one file.  Overall, I don't think it should be too big of a performance hit though and to be honest, performance is not my biggest concern on a script that reboots the server multiple times.

cPholloway:
^ this is just a rebase on main

toddr (approved):

cPholloway:
^ rebase on main

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,461 @@
...
+            return;
+        }
+
+        # If the service was not up at this point, it is likely that
+        # the schema update failed during the install.  In this case,
+        # the only way I have found to get things working again is to
+        # completely remove the package and start the install over
+
+        $self->remove_cpanel_ccs_home_directory();
+        $self->remove_cpanel_ccs_home_directory();
```
troglodyne, line 378: Redundant?

lib/Elevate/Components/CCS.pm
```diff
@@ -0,0 +1,461 @@
...
+        $self->_clear_task_queue();
+        $self->_install_ccs_and_dependencies();
```
troglodyne, line 380: Need to run reinstall here?

lib/Elevate/Components/CCS.pm
cPholloway: Sigh, copy/paste error.  Line 377 should be removing the package

tbordner (approved):
Latest commits seem to have done the trick and I haven't encountered service errors while using this since.

PR #413, opened 2024-04-11
Ensure all system calls are executable absolute paths

cPholloway:
Case RE-260: C::S::O has a bug in 110 where it does not detect missing scripts (see CPANEL-40343).  The change in CPANEL-40343 was deemed medium/high risk, and was rejected as a backport candidate for 110.  As such, we need to work around this issue.  Since elevate-cpanel can only be executed as root, it was determined that all commands that it executes should be executabl absolute paths.  This change enforces that.

Changelog:  Ensure all system calls are executable absolute paths

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -5609,6 +5609,9 @@ EOS
 
     sub _ssystem ( $command, %opts ) {
         my @args = @{ $command // [] };
+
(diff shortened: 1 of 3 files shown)


tbordner (approved):
QA looks good, workaround appears to be safe!

annes449 (approved):
QA is all good.

toddr (approved):

PR #417, opened 2024-04-18
Move logic for prepping cPanel for leapp to a component.

sfbee:
Case RE-122: This change moves logic meant to prep cPanel services for the leapp into a component, adds a call to flush the cPanel task queue before leapping, and adds unit test coverage for all of the above.

Changelog: Move logic for prepping cPanel services to leapp into a component, and flush task queue before leapp.


timmullin:
I think we need to rebase this off of the latest in the main branch.  When I look at the changes in `scripts/elevate-cpanel.PL` I'm not sure whether the line containing `$self->run_component_once( 'cPanelPrep', => 'pre_leapp' );` will end up before or after `$self->run_component_once( 'DatabaseUpgrade' => 'pre_leapp' );`; and it needs to come after.

lib/Elevate/Components/cPanelPrep.pm
```diff
@@ -0,0 +1,114 @@
...
+use Elevate::SystemctlService ();
+
+use Cpanel::FileUtils::TouchFile ();
+
+use Cwd           ();
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Components::Base};
+
+sub pre_leapp ($self) {
```
timmullin, line 29: I wonder if we should also have a post_leapp() function that could undo some of the prep, like re-enabling the services, etc.

cPholloway (approved):

annes449 (approved):
QA is all good.

PR #420, opened 2024-04-22
Do not delay 10 minutes when elevate_leap_fail_continue is put in place

toddr:
Case RE-305: Also restructured _wait_for_log_contents to handle corner cases for a better flow


annes449 (approved):
QA is all good.

PR #422, opened 2024-04-23
Fix DNS name server blocker

timmullin:
Case RE-318:  Updated to check name server type the system is using in order to determine whether to block on it.

Changelog: Fix DNS blocker to check the systems name server type.


toddr (approved):

tbordner (approved):
Seemed to check now as fully intended.

annes449 (approved):
QA looks good.

PR #427, opened 2024-04-26
Do not block on dropped PHP packages if no domains use them

cPholloway:
Case RE-313:  Currently, the EA4 blocker will block if there are any packages that will be dropped (not provided on the newer OS) during the elevation.  This change teaches the blocker to check to see if any domains are using the installed version of PHP.  If there are no domains using the version of PHP that will be dropped, then we no longer block for that version of PHP and allow the elevation to continue if that was the only thing blocking it.

Changelog:

### DIFF
--- a/lib/Elevate/Blockers/EA4.pm
+++ b/lib/Elevate/Blockers/EA4.pm
@@ -15,6 +15,9 @@ use cPstrict;
 use Elevate::EA4       ();
 use Elevate::StageFile ();
 
+use Cpanel::JSON            ();
+use Cpanel::SafeRun::Simple ();
+
 use parent qw{Elevate::Blockers::Base};
 
 use Cwd           ();
@@ -41,17 +44,7 @@ sub _blocker_ea4_profile ($self) {
 
     Elevate::EA4::backup();
 
-    my $stash        = Elevate::StageFile::read_stage_file();
-    my $dropped_pkgs = $stash->{'ea4'}->{'dropped_pkgs'} // {};
-    return unless scalar keys $dropped_pkgs->%*;
-
-    my @incompatible_packages;
-    foreach my $pkg ( sort keys $dropped_pkgs->%* ) {
-        my $type = $dropped_pkgs->{$pkg} // '';
-        next if $type eq 'exp';                          # use of experimental packages is a non blocker
-        next if $pkg =~ m/^ea-openssl(?:11)?-devel$/;    # ignore these packages, as they can be orphans
-        push @incompatible_packages, $pkg;
-    }
+    my @incompatible_packages = $self->_get_incompatible_packages();
 
     return unless @incompatible_packages;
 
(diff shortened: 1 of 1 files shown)


timmullin:
This pull request is getting this particular failure:
`not ok 50 - Elevate::EA4 is fatpack'd in elevate-cpanel script`

lib/Elevate/Blockers/EA4.pm
```diff
@@ -41,27 +44,73 @@ sub _blocker_ea4_profile ($self) {
...
     }
 
-    return unless @incompatible_packages;
+    return @incompatible;
+}
 
-    my $list = join( "\n", map { "- $_" } @incompatible_packages );
+sub _php_version_is_in_use ( $self, $php ) {
+    my $current_php_usage = $self->_get_php_versions_in_use();
+    return $current_php_usage->{$php} ? 1 : 0;
```
timmullin, line 85: We need to return true if `api_fail` is set in order to assume it is in use so we can block as according to the warning we give if the API fails.

lib/Elevate/Blockers/EA4.pm
```diff
@@ -41,27 +44,73 @@ sub _blocker_ea4_profile ($self) {
...
+    unless ( $result->{metadata}{result} ) {
+
+        WARN( <<~"EOS" );
+        Unable to determine if PHP versions that will be dropped are in use by
+        a domain.  Assuming that they are in use and blocking to be safe.
+
+        EOS
+
+        $php_versions_in_use->{api_fail} = 1;
+        return;
```
timmullin, line 105: Returning undef will cause an error on line 85.  If we want to assume all PHP versions are in use when the API call fails, we need to return `$php_versions_in_use` so the caller can test if `api_fail` is set.

cPholloway:
> This pull request is getting this particular failure: `not ok 50 - Elevate::EA4 is fatpack'd in elevate-cpanel script`

I'm not too surprised that the CI is failing on this.  I did not update `elevate-cpanel` in this PR since this is going to a feature branch.  I will ensure the CI passes when I work on RE-330

lib/Elevate/Blockers/EA4.pm
```diff
@@ -41,27 +44,77 @@ sub _blocker_ea4_profile ($self) {
...
+
+    unless ( $result->{metadata}{result} ) {
+
+        WARN( <<~"EOS" );
+        Unable to determine if PHP versions that will be dropped are in use by
+        a domain.  Assuming that they are in use and blocking to be safe.
+
+        EOS
+
+        $php_versions_in_use->{api_fail} = 1;
```
toddr, line 108: Wouldn't it be better to just assume they're all in use if we can't get help from the API? It would be useful for this to feed back into our stats but ideally not in a blocking way.

toddr (approved):

PR #429, opened 2024-05-02
Add blocker if '/usr' is a separate private mount point

cPholloway:
Case RE-306:  The 'leapp' utility needs to be able to load packages stored within the
'/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/files/bundled-rpms' directory.  If '/usr' is a separate private bind mount, the 'leapp' utility will be unable to load the packages stored within this directory which will result in 'leapp' failing, and leaving systems in a broken state.  In order to avoid this, we now detect this scenario and block the elevation.

Changelog: Add blocker if '/usr' is a separaate private mount point


toddr (approved):

annes449 (approved):
QA tests pass.

PR #430, opened 2024-05-07
Teach EA4 blocker about hardened PHP provided by Imunify 360

cPholloway:
Case RE-313: Imunify 360 is a paid product that provides hardened PHP versions that are available on newer OS versions.  Currently, the EA4 blocker would block on older PHP versions such as ea-php54 since they are EOL and not provided on AlmaLinux 8 by cPanel.  However, Imunify 360 provides these versions of PHP on AlmaLlinux 8 as well as 7 so elevate should not block if the hardened PHP version is provided by Imunify 360. This change updates the EA4 blocker to no longer block on hardened PHP versions that are provided by Imunify 360.

Changelog:

### DIFF
--- a/lib/Elevate/Blockers/EA4.pm
+++ b/lib/Elevate/Blockers/EA4.pm
@@ -12,10 +12,12 @@ Blocker to check EasyApache profile compatibility.
 
 use cPstrict;
 
+use Elevate::Constants ();
 use Elevate::EA4       ();
 use Elevate::StageFile ();
 
 use Cpanel::JSON            ();
+use Cpanel::Pkgr            ();
 use Cpanel::SafeRun::Simple ();
 
 use parent qw{Elevate::Blockers::Base};
@@ -64,6 +66,7 @@ sub _get_incompatible_packages ($self) {
     return unless scalar keys $dropped_pkgs->%*;
 
     my @incompatible;
+    my @imunify_pkgs;
     foreach my $pkg ( sort keys %$dropped_pkgs ) {
         my $type = $dropped_pkgs->{$pkg} // '';
         next if $type eq 'exp';                          # use of experimental packages is a non blocker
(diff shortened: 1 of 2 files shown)


lib/Elevate/Blockers/EA4.pm
```diff
@@ -72,14 +75,33 @@ sub _get_incompatible_packages ($self) {
...
 
+    if (@imunify_pkgs) {
+        Elevate::StageFile::update_stage_file( { ea4_imunify_packages => \@imunify_pkgs } );
+    }
+
     return @incompatible;
 }
 
+sub _php_is_provided_by_imunify_360 ( $self, $php ) {
+    return 0 unless -x Elevate::Constants::IMUNIFY_AGENT;
```
toddr, line 96: this is only there if 360 is present?

lib/Elevate/Components/EA4.pm
```diff
@@ -149,4 +150,13 @@ sub _restore_config_files ($self) {
     return;
 }
 
+sub _restore_imunify_phps ($self) {
+
+    my $php_pkgs = Elevate::StageFile::read_stage_file('ea4_imunify_packages');
+    return unless scalar $php_pkgs->@*;
+
+    $self->dnf->install(@$php_pkgs);
```
toddr, line 158: so we're using yum after we restore EA? this means EA4 restore will fail and we're ok with that?

lib/Elevate/Blockers/EA4.pm
cPholloway: I'm not sure about this, but I think it might be any Imunify product.  This was just what `Elevate::Components::Imunify::is_installed` was using to determine if it was installed, so I used it here.  It should not hurt anything if we let things through with regular Imunify installed, but I could look for a better way to make sure it is Imunify 360 specifically if preferred.  Thoughts?

lib/Elevate/Components/EA4.pm
cPholloway: The EA4 restore will be successful, but it will not install these packages since the backup profile considers them to be dropped packages.

Doing it this way came down to a design decision:

1.  Manually manipulate the profile file that `ea_current_to_profile` generates so that the imunify packages get reinstalled

or

2.  Track the "dropped" packages from Imunify 360 that EA4 doesn't know about, then install them after the EA4 restore completes

I went with option 2 here because I felt that it was safer and easier to proceed this way than with option 1.

PR #432, opened 2024-05-08
Enhance EA4 blocker

cPholloway:
Case RE-313:  Currently, the EA4 blocker is biggest road blocks for our customers to being able to elevate their servers to A8/CL8.  This change enhances the EA4 blocker to no longer block on installed versions of PHP that are not provided by A8/CL8 unless there are domains actively using the installed version of PHP.  It also teaches the EA4 blocker about hardened PHP versions that are provided by the Imunify 360 plugin since these versions of PHP would are available on A8.  Additionally, it refactors some logic that is shared between the EA4 blocker and EA4 component into a shared library (Elevate::EA4).

Changelog:  Enhance EA4 blocker so that it only blocks on installed PHP
 versions that are not provided by Imunify 360 and are actively in use by
 domains on the server


annes449 (approved):
All good with QA.

toddr (approved):

PR #434, opened 2024-05-10
Have script block if mount -a fails to exit cleanly

cPholloway:
Case RE-134: Since the script reboots the server and relies on the file system to be consistent between reboots, we need to ensure that mount -a exits cleanly before starting the elevate process.  This change adds a blocker to the script is mount -a fails to exit cleanly.

Changelog: Have script block if mount -a fails to exit cleanly

### DIFF
--- a/docs-website-src/content/blockers.md
+++ b/docs-website-src/content/blockers.md
@@ -74,6 +74,10 @@ You can discover many of these issues by downloading `elevate-cpanel` and runnin
 
 The following is a list of other known issues that could prevent your server's successful elevation.
 
+## Filesystem
+
+Since elevate needs to reboot your system multiple times as part of the upgrade process, we ensure that the command 'mount -a' succeeds successfully before allowing the elevation to proceed.  The reason for this is that we need to be able to trust that the filesystem remains the same between each reboot.
+
 ## PostgreSQL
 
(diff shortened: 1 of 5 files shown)


timmullin (approved):

lib/Elevate/Blockers/MountPoints.pm
```diff
@@ -15,10 +15,53 @@ use cPstrict;
...
 sub check ($self) {
+    $self->_check_for_rhel_23449();
+    $self->_ensure_mount_dash_a_succeeds();
+    return;
+}
+
+sub _ensure_mount_dash_a_succeeds ($self) {
+
+    # Only do this in start mode because it can change the file system mounts
+    return if $self->is_check_mode();
```
toddr, line 31: Is there a reason we don't want to do this in check? Maybe we discussed it and I've slept since then.

toddr: Never mind

toddr (approved):

annes449 (approved):
All good with QA.

docs-website-src/content/blockers.md
```diff
@@ -74,6 +74,10 @@ You can discover many of these issues by downloading `elevate-cpanel` and runnin
 
 The following is a list of other known issues that could prevent your server's successful elevation.
 
+## Filesystem
+
+Since elevate needs to reboot your system multiple times as part of the upgrade process, we ensure that the command 'mount -a' succeeds successfully before allowing the elevation to proceed.  The reason for this is that we need to be able to trust that the filesystem remains the same between each reboot.
```
cPanelStacyF, line 79: Other instances of "elevate" in this document are written as ELevate, this one should match. I'd also make it read: the `mount -a` command

PR #439, opened 2024-05-15
Unblock remote MySQL blocker by temporarily disabling during upgrade.

sfbee:
Case RE-34:

Changelog: Unblock remote MySQL blocker by temporarily disabling the
 remote profile during the upgrade.

### DIFF
--- a/lib/Elevate/Components/DatabaseUpgrade.pm
+++ b/lib/Elevate/Components/DatabaseUpgrade.pm
@@ -1,7 +1,6 @@
 package Elevate::Components::DatabaseUpgrade;
 
 =encoding utf-8
-
 =head1 NAME
 
 Elevate::Components::DatabaseUpgrade
@@ -16,6 +15,20 @@ use Elevate::Database ();
 
 use parent qw{Elevate::Components::Base};
 
+use Cpanel::MysqlUtils::RemoteMySQL::ProfileManager ();
+
+use File::Slurper;
+use Try::Tiny;
+
+use Cpanel::MysqlUtils::MyCnf::Basic                ();
+use Cpanel::MysqlUtils::RemoteMySQL::ProfileManager ();
+use Cpanel::PasswdStrength::Generate                ();
+use Cpanel::JSON                                    ();
+use Cpanel::SafeRun::Simple                         ();
+use Cpanel::Encoder::URI                            ();
+
+use constant MYSQL_PROFILE_FILE => '/var/cpanel/elevate-mysql-profile';
+
 use Log::Log4perl qw(:easy);
 
 sub pre_leapp ($self) {
@@ -26,6 +39,8 @@ sub pre_leapp ($self) {
     # If the database version is supported on the new OS version, then no need to upgrade
     return if Elevate::Database::is_database_version_supported( Elevate::Database::get_local_database_version() );
 
+    $self->_ensure_localhost_mysql_profile_is_active(1);
+
     Elevate::Database::upgrade_database_server();
 
     return;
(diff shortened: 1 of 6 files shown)


toddr:
@sfbee please remove cover_db from this commit.

lib/Elevate/Components/DatabaseUpgrade.pm
```diff
@@ -26,15 +38,95 @@ sub pre_leapp ($self) {
...
+        }
+    };
+
+    return;
+}
+
+sub _create_new_localhost_profile {
+    my ($profile_manager) = @_;
+
+    my $active_profile = $profile_manager->get_active_profile('dont_die');
```
cPholloway, line 94: FIX:  We should `try/catch` this and die with an elevate specific failure message if we can't get the active profile.  Otherwise, `$active_profile` is undef on line 95 and we do can't write out the active profile on line 95.  Due to this, we will not be able to switch the active profile back to the remote server post elevate either.

lib/Elevate/Components/DatabaseUpgrade.pm
```diff
@@ -26,15 +38,95 @@ sub pre_leapp ($self) {
...
+    File::Slurper::write_text( MYSQL_PROFILE_FILE, $active_profile );
+
+    my $password = Cpanel::PasswdStrength::Generate::generate_password( 16, no_othersymbols => 1 );
+
+    $profile_manager->create_profile(
```
cPholloway, line 99: FIX:  We need to `try/catch` this and die with an elevate specific message if we can't create a locahost profile here.  Otherwise, it will die with an uncaught exception if it fails to create a localhost profile

lib/Elevate/Components/DatabaseUpgrade.pm
```diff
@@ -26,15 +38,95 @@ sub pre_leapp ($self) {
...
+            'mysql_pass' => $password,
+            'mysql_host' => 'localhost',
+            'mysql_port' => Cpanel::MysqlUtils::MyCnf::Basic::getmydbport('root') || 3306,
+            'setup_via'  => 'Auto-generated localhost profile during elevate.',
+        },
+        { 'overwrite' => 1 },
+    );
+
+    INFO("Resetting password for local root MySQL user...");
+    Cpanel::SafeRun::Simple::saferunnoerror( q{/bin/sh}, q{-c}, qq{/usr/local/cpanel/bin/whmapi1 set_local_mysql_root_password password='$password'} );
```
cPholloway, line 112: FIX:  We should capture the output of the API call and make sure that it succeeds.  If the API call fails, we either need to retry or die here.

cPholloway (changes requested):

lib/Elevate/Components/DatabaseUpgrade.pm
```diff
@@ -26,15 +38,95 @@ sub pre_leapp ($self) {
...
+    if ( Cpanel::MysqlUtils::MyCnf::Basic::is_local_mysql() ) {
+        return 1;
+    }
+
+    my $profile_manager = Cpanel::MysqlUtils::RemoteMySQL::ProfileManager->new();
+
+    # Validate that the current “localhost” profile exists, and contains valid settings.
+    try {
+        $profile_manager->validate_profile('localhost');
+        _activate_localhost_profile($profile_manager);
```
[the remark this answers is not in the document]
sfbee: validate_profile() dies if the profile isn't valid, and _activate_localhost_profile() dies if it can't close out the profile manager's lock

cPholloway:
FIX:

1.  This branch is 18 commits behind main to make sure there are no conflicts.
2. I am not seeing any new tests in this commit.  By change, did those not get added in with this commit?

cPholloway (changes requested):

t/components-DatabaseUpgrade.t
```diff
@@ -0,0 +1,258 @@
...
+use Test::MockModule qw/strict/;
+use Test::MockFile   qw/strict/;
+
+use lib $FindBin::Bin . "/lib";
+use Test::Elevate;
+
+use cPstrict;
+
+use strict;
+use warnings;
```
cPholloway, line 26: TRIVIAL:  We do not need these since this uses `cPstrict`

cPholloway (approved):

lib/Elevate/Components/DatabaseUpgrade.pm
```diff
@@ -16,6 +15,20 @@ use Elevate::Database ();
...
+use Try::Tiny;
+
+use Cpanel::MysqlUtils::MyCnf::Basic                ();
+use Cpanel::MysqlUtils::RemoteMySQL::ProfileManager ();
+use Cpanel::PasswdStrength::Generate                ();
+use Cpanel::JSON                                    ();
+use Cpanel::SafeRun::Simple                         ();
+use Cpanel::Encoder::URI                            ();
+
+use constant MYSQL_PROFILE_FILE => '/var/cpanel/elevate-mysql-profile';
```
tbordner, line 30: As far as I can tell, this isn't being created? At no point has Elevate tried to restore the remote instances after this whether they are still valid or not.

tbordner (changes requested):
The blocker is removed, elevation succeeds, and the remote MySQL configurations are preserved but NOT reactivated after the update.

No configurations seem to trigger the path starting at line 5278.

toddr:
I'm declining this request because it needs to be submitted by someone else at this point.

PR #440, opened 2024-05-16
Have pre_leapp component remove packages that leapp will remove

cPholloway:
Case RE-132:  The leapp utility will remove some packages due to incompatibility/dependency related issues after the upgrade to AlmaLinux/CloudLinux 8 has been completed.  Rather than relying on leapp to handle the removal of these packages (potential upstream bugs), we should instead manually remove them as part of a pre_leapp component. This change updates a component to perform this removal prior to leapp executing.

Changelog:  Have pre_leapp component remove packages that leapp will
 remove

### DIFF
--- a/lib/Elevate/Components/RpmDB.pm
+++ b/lib/Elevate/Components/RpmDB.pm
@@ -24,6 +24,17 @@ use Cpanel::Yum::Vars ();
 
 use parent qw{Elevate::Components::Base};
 
+use constant OBSOLETE_PACKAGES => (
+    'compat-db',
+    'gd-progs',
+    'python-tools',
+    'python2-dnf',
+    'python2-libcomps',
+    'tcp_wrappers-devel',
+    'tkinter',
+    'yum-plugin-universal-hooks',
+);
+
 sub pre_leapp ($self) {
 
     $self->run_once("_cleanup_rpms");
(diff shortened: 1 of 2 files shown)


lib/Elevate/Components/RpmDB.pm
```diff
@@ -24,6 +24,17 @@ use Cpanel::Yum::Vars ();
...
 use parent qw{Elevate::Components::Base};
 
+use constant OBSOLETE_PACKAGES => (
+    'compat-db',
+    'gd-progs',
+    'python-tools',
+    'python2-dnf',
+    'python2-libcomps',
+    'tcp_wrappers-devel',
+    'tkinter',
```
toddr, line 34: I'm ok with rpm -e --nodeps but this will actually remove things that may have side effects.
```
Removing:
 tkinter                                                 x86_64                                  2.7.5-94.el7_9                                        @updates                                   1.2 M
Removing for dependencies:
 cpanel-dnspython                                        x86_64                                  1.12.0-4.cp108~el7                                    installed                                  960 k
 cpanel-mailman                                          x86_64                                  2.1.39-1.cp108~el7                                    installed                                   32 M
 cpanel-pythontidy                                       noarch                                  1.22-1.cp108~el7                                      installed                                  144 k
 cpanel-system-python27                                  x86_64                                  2.7.18-1.cp108~el7                                    installed                                  0.0  
 python-tools                                            x86_64                                  2.7.5-94.el7_9                                        @updates                                   2.8 M

```

cPholloway: This is executing just before leapp is installed and leapp upgrade is executed.  At this point, all of the cPanel RPMs have already been removed, and leapp will remove these packages like this anyway.  When I ran through the happy path with this in place, I did not see any side affects from doing this.

PR #441, opened 2024-05-17
Add blocker for an invalid Imunify license

cPholloway:
Case RE-373: If a server has an invalid Imunify license, then the Imunify component can fail in expected ways.  This can further lead to the elevate script failing in stage 4 due to the Imunify post_leapp component failing and the EA4 component relying on the Imunify component to succeed.

Changelog: Add blocker for an invalid Imunify license

### DIFF
--- a/lib/Elevate/Blockers/Imunify.pm
+++ b/lib/Elevate/Blockers/Imunify.pm
@@ -1,0 +1,48 @@
+package Elevate::Blockers::Imunify;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Blockers::Imunify
+
+Blocker to check that the Imunify license is valid when Imunify is installed.
+
+=cut
+
+use cPstrict;
+
+use Elevate::Constants ();
+
+use Cpanel::JSON ();
+
+use parent qw{Elevate::Blockers::Base};
+
+use Log::Log4perl qw(:easy);
+
+sub check ($self) {
+    return $self->_check_imunify_license();
+}
+
+sub _check_imunify_license ($self) {
+    return unless -x Elevate::Constants::IMUNIFY_AGENT;
+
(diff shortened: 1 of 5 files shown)


lib/Elevate/Blockers/Imunify.pm
```diff
@@ -0,0 +1,50 @@
...
+    return unless -x Elevate::Constants::IMUNIFY_AGENT;
+
+    my $agent_bin    = Elevate::Constants::IMUNIFY_AGENT;
+    my $out          = $self->ssystem_capture_output( $agent_bin, 'version', '--json' );
+    my $raw_data     = join "\n", @{ $out->{stdout} };
+    my $license_data = eval { Cpanel::JSON::Load($raw_data) } // {};
+
+    if ( !ref $license_data->{license} || !$license_data->{license}->{status} ) {
+
+        $self->blockers->abort_on_first_blocker(1);
```
toddr, line 37: Won't this block our ability to get --check data from the other blockers? I'm not sure why we'd abort on this one?

toddr (changes requested):
I don't think we need to abort on this blocker aggressively.

lib/Elevate/Blockers/Imunify.pm
cPholloway: Yes, it will block our ability to get data from other blockers.  I went ahead and aborted on this one since we also abort on the CloudLinux license blocker.

cPholloway: I'll got ahead and remove this line though as it is probably not necessary

timmullin (approved):

toddr (approved):

annes449 (approved):
All good with QA

PR #449, opened 2024-06-05
No longer block on unfinished yum transactions

timmullin:
Case RE-338:

But, when performing the upgrade, go ahead and clean up the unfinished yum transactions.

Changelog: No longer blocks if there are unfinished yum transactions.

### DIFF
--- a/lib/Elevate/Blockers/Repositories.pm
+++ b/lib/Elevate/Blockers/Repositories.pm
@@ -23,10 +23,13 @@ use parent qw{Elevate::Blockers::Base};
 
 use Log::Log4perl qw(:easy);
 
+use constant YUM_COMPLETE_TRANSACTION_BIN => '/usr/sbin/yum-complete-transaction';
+use constant FIX_RPM_SCRIPT               => '/usr/local/cpanel/scripts/find_and_fix_rpm_issues';
+
 sub check ($self) {
     my $ok = 1;
-    $ok = 0 unless $self->_blocker_invalid_yum_repos;
-    $ok = 0 unless $self->_yum_is_stable();
+    $ok = 0 if $self->_blocker_invalid_yum_repos;
+    $ok = 0 if $self->_yum_is_stable();
 
     return $ok;
 }
@@ -104,7 +107,7 @@ sub _yum_is_stable ($self) {
             ERROR('yum appears to be unstable. Please address this before upgrading');
             ERROR($errors);
             my $id = ref($self) . '::YumMakeCacheError';
-            $self->has_blocker(
+            return $self->has_blocker(
                 "yum appears to be unstable. Please address this before upgrading\n$errors",
                 info => {
                     name  => $id,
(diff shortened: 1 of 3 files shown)


lib/Elevate/Blockers/Repositories.pm
```diff
@@ -121,21 +121,18 @@ sub _yum_is_stable ($self) {
...
-                    error        => 'YUM has unfinished transactions',
-                    transactions => join( "\n", sort @transactions ),
-                },
-                blocker_id => $id,
-                quiet      => 1,
-            );
-
-            return 0;
+            if ( $self->is_check_mode() ) {
+                INFO('We will attempt to clear the unfinished yum transactions upon upgrade.');
```
toddr, line 127: Maybe add the message from the old blocker so they know?

toddr: The tool `yum-complete-transaction` may help you repair this in advance if you prefer to review it.

toddr (changes requested):
Minor strings change requested. OK to QA

lib/Elevate/Blockers/Repositories.pm
toddr: WARN('Unfinished yum transactions detected. yum-complete-transactions will display these issues. Otherwise, elevate will execute /usr/sbin/yum-complete-transaction --cleanup-only on start');

lib/Elevate/Blockers/Repositories.pm
```diff
@@ -121,21 +121,18 @@ sub _yum_is_stable ($self) {
...
-
-            return 0;
+            if ( $self->is_check_mode() ) {
+                WARN('Unfinished yum transactions detected. Elevate will execute /usr/sbin/yum-complete-transaction --cleanup-only during upgrade');
+            }
+            else {
+                INFO('Cleaning up unfinished yum transactions.');
+                my $ret = $self->ssystem_capture_output(qw{/usr/sbin/yum-complete-transaction --cleanup-only});
+                if ( $ret->{status} != 0 ) {
+                    WARN( "Errors encountered running \"yum-complete-transaction\": " . $ret->{stderr} );
```
cPholloway, line 133: We should die or throw a blocker since we are still in the blocker file if `yum-complete-transaction` fails.  This could be an indication of a much larger problem if it does fail.

cPholloway (changes requested):

lib/Elevate/Blockers/Repositories.pm
timmullin: It could be that they just didn't have yum-utils installed.  I didn't think a problem with that utility itself should be a show-stopper.  Especially since leapp doesn't really care about unfinished yum transactions.

cPholloway: Fair enough, I'll go ahead and approve this.

cPholloway (approved):

cPholloway (approved):

annes449 (approved):
QA is all good.

PR #451, opened 2024-06-10
Remove ELS for CentOS 7

timmullin:
Case RE-368:

Use the manual method to remove ELS. This involves deleting two files and un-installing the package.

Changelog: Removes ELS for CentOS 7.

### DIFF
--- a/lib/Elevate/Components/ELS.pm
+++ b/lib/Elevate/Components/ELS.pm
@@ -1,0 +1,52 @@
+package Elevate::Components::ELS;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::ELS
+
+Remove ELS for CentOS 7
+
+=cut
+
+use cPstrict;
+
+use Elevate::OS ();
+
+use Cpanel::Pkgr  ();
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Components::Base};
+
+use constant ELS_PACKAGE => 'els-define';
+
+sub pre_leapp ($self) {
+
+    return unless Elevate::OS::remove_els();
+
+    return unless Cpanel::Pkgr::is_installed(ELS_PACKAGE);
+
+    my @files_to_remove = qw{
+      /etc/yum.repos.d/centos7-els.repo
+      /etc/yum.repos.d/centos7-els-rollout.repo
+    };
+
+    foreach my $file (@files_to_remove) {
+        if ( -e $file ) {
+            unlink $file or WARN("Could not remove file $file: $!");
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


lib/Elevate/Components/ELS.pm
```diff
@@ -0,0 +1,53 @@
...
+      /etc/yum.repos.d/centos7-els-rollout.repo
+    };
+
+    foreach my $file (@files_to_remove) {
+        if ( -e $file ) {
+            unlink $file or WARN("Could not remove file $file: $!");
+        }
+    }
+
+    $self->yum->remove(ELS_PACKAGE);
```
toddr, line 42: This is my best guess too but I'm uncertain if there's more than one package to remove :\

toddr (approved):

lib/Elevate/Components/ELS.pm
timmullin: I got that from the manual steps outlined in the document referenced in the case.

cPholloway: The rest is replacing system packages.  My guess is that leapp handles replacing the rest with upgrades for A8.

cPholloway (approved):

lib/Elevate/Components/ELS.pm
```diff
@@ -0,0 +1,53 @@
...
+use constant ELS_PACKAGE => 'els-define';
+
+sub pre_leapp ($self) {
+
+    return unless Elevate::OS::remove_els();
+
+    return unless Cpanel::Pkgr::is_installed(ELS_PACKAGE);
+
+    my @files_to_remove = qw{
+      /etc/yum/vars/elstoken
```
cPholloway, line 31: @timmullin - I just got an updated version of the install script for ELS and with IP based licensing, this file is not needed.  Can you remove it from the list of files to remove please?

timmullin: I can go ahead and remove that.

timmullin: Now that I think about it, what if els had been installed with the older version of the script and the file is present?  Plus, we check if the file exists before deleting it; so, the file's absence would not create an error condition for the removal of els if they had used the newer version of the script.  It doesn't seem that there is any harm in leaving it in.  And, there are probably plenty of installs out there that have used the older version of the script and ought to have that file removed.

cPholloway (approved):

lib/Elevate/Components/ELS.pm
toddr: agreed!

annes449 (approved):
QA is good!

PR #453, opened 2024-06-14
Case RE-361: Make it more clear why yum makecache failures are a problem

davelcpanelnet:
Add some simple verbiage in that shows in all cases given `yum makecache` is non-zero.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1320,7 +1320,7 @@ EOS
     The grub2-pc package is not installed. The GRUB2 boot loader is
     required to upgrade via leapp.
 
-    If you need assistance, open a ticket with cPanel Support, as outlined here:
+    You may want to consider reaching out to cPanel Support for assistance:
     https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
     EOS
     }
@@ -1333,7 +1333,7 @@ EOS
             return $self->has_blocker( <<~EOS );
         The GRUB2 config file is missing.
 
-        If you need assistance, open a ticket with cPanel Support, as outlined here:
+        You may want to consider reaching out to cPanel Support for assistance:
         https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
         EOS
         }
@@ -1983,12 +1983,12 @@ EOS
         if ( $errors =~ m/\S/ms ) {
 
             my $error_msg = <<~'EOS';
-            '/usr/bin/yum makecache' failed to return cleanly. This could be due to a temporary mirror problem, or it could indicate a larger issue, such as a broken repository. Since this script relies heavily on yum, you will need to address this issue before upgrading.
+        '/usr/bin/yum makecache' failed to return cleanly. This could be due to a temporary mirror problem, or it could indicate a larger issue, such as a broken repository. Since this script relies heavily on yum, you will need to address this issue before upgrading.
 
-            If you need assistance, open a ticket with cPanel Support, as outlined here:
+        If you need assistance, open a ticket with cPanel Support, as outlined here:
 
-            https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
-            EOS
+        https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
+        EOS
 
             WARN("Initial run of \"yum makecache\" failed: $errors");
             WARN("Running \"yum clean all\" in an attempt to fix yum");
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1320,7 +1320,7 @@ EOS
     The grub2-pc package is not installed. The GRUB2 boot loader is
     required to upgrade via leapp.
 
-    You may want to consider reaching out to cPanel Support for assistance:
+    If you need assistance, open a ticket with cPanel Support, as outlined here
     https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
     EOS
     }
@@ -1333,7 +1333,7 @@ EOS
             return $self->has_blocker( <<~EOS );
         The GRUB2 config file is missing.
 
-        You may want to consider reaching out to cPanel Support for assistance:
+        If you need assistance, open a ticket with cPanel Support, as outlined here
         https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
         EOS
         }
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1982,6 +1982,14 @@ EOS
         my $errors = Cpanel::SafeRun::Errors::saferunonlyerrors(qw{/usr/bin/yum makecache});
         if ( $errors =~ m/\S/ms ) {
 
+            my $error_msg = <<~'EOS';
+            '/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.
+
+            You may want to consider reaching out to cPanel Support for assistance:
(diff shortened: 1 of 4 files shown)


lib/Elevate/Blockers/Repositories.pm
```diff
@@ -104,11 +104,11 @@ sub _yum_is_stable ($self) {
 
         $errors = Cpanel::SafeRun::Errors::saferunonlyerrors(qw{/usr/bin/yum makecache});
         if ( $errors =~ m/\S/ms ) {
-            ERROR('yum appears to be unstable. Please address this before upgrading');
+            ERROR('yum appears to be unstable. Specifically, "yum makecache" is generating an error.  Please address this before upgrading');
```
cPholloway, line 107: Lets expand this out a bit further.  Maybe something like this:

```
my $error_msg = <<~'EOS';
'/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.

You may want to consider reaching out to cPanel Support for assistance:

https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/

EOS
```

Also, please execute `make` before resubbing this so that the fatpacker will update the main `elevate-cpanel` script

cPholloway (changes requested):

lib/Elevate/Blockers/Repositories.pm
davelcpanelnet: Cool.  Easily done!

elevate-cpanel
```diff
@@ -1982,6 +1982,15 @@ EOS
         my $errors = Cpanel::SafeRun::Errors::saferunonlyerrors(qw{/usr/bin/yum makecache});
         if ( $errors =~ m/\S/ms ) {
 
+        my $error_msg = <<~'EOS';
+        '/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.
+
+        You may want to consider reaching out to cPanel Support for assistance:
```
toddr, line 1988: I'm concerned with this advice. This advice could be supplied everywhere in theory. why here?

cPholloway: We already recommend this in a few spots:

```
# git grep 'how-to-open-a-technical-support-ticket' lib/ script/
lib/Elevate/Blockers/Grub2.pm:        https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
lib/Elevate/Blockers/Grub2.pm:    https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
lib/Elevate/Blockers/Grub2.pm:        https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
lib/Elevate/Blockers/MountPoints.pm:        https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
lib/Elevate/Components/CCS.pm:        https://docs.cpanel.net/knowledge-base/technical-support-services/how-to-open-a-technical-support-ticket/
```

I thought it made sense to do it in this instance because there are lots of different ways that `yum makecache` can fail and there is not a good automated way to determine why it failed due to the number of errors that could be emitted from it.  It also made sense to me here because support deals with yum issues, including makecache failing, quite frequently due to EA4 among other things so they should be well equiped to handle issues that come in due to this.

cPholloway (approved):

lib/Elevate/Blockers/Repositories.pm
```diff
@@ -94,6 +94,14 @@ sub _yum_is_stable ($self) {
     my $errors = Cpanel::SafeRun::Errors::saferunonlyerrors(qw{/usr/bin/yum makecache});
     if ( $errors =~ m/\S/ms ) {
```
toddr, line 96: messge

toddr (approved):

elevate-cpanel
davelcpanelnet: What @cPholloway was getting at was more or less my take.

If we wanted to extend some method to be more useful we would probably need to identify something like the top x known errors after parsing through tickets to identify what common errors were found - then implementing some sort of regex to trigger whatever text advise etc we would want to display if one of those common errors was identified.

Past that we could just look for core patterns like:

- /Resolving Dependencies/&&/Requires:\s/ - could obviously show a dependency issue taking place and recommend filing a ticket
- /Error: requested datatype/ - could signify a bad repo is sitting stale in /etc/yum.repos.d/*.repo

This just seemed the simplest thing for the surface area.

elevate-cpanel
```diff
@@ -1982,6 +1982,14 @@ EOS
         my $errors = Cpanel::SafeRun::Errors::saferunonlyerrors(qw{/usr/bin/yum makecache});
         if ( $errors =~ m/\S/ms ) {
 
+            my $error_msg = <<~'EOS';
+            '/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.
```
sbillik, line 1986: - Modern protocol uses only a single space after a period.
- Missing commas
- Passive voice
```suggestion
            '/usr/bin/yum makecache' failed to return cleanly. This could be due to a temporary mirror problem, or it could indicate a larger issue, such as a broken repository. Since this script relies heavily on yum, you will need to address this issue before upgrading.
```

elevate-cpanel
```diff
@@ -1982,6 +1982,14 @@ EOS
...
+
+            You may want to consider reaching out to cPanel Support for assistance:
```
sbillik, line 1988: This doesn't seem strong enough.

```suggestion
            If you need assistance, open a ticket with cPanel Support, as outlined here:
```

lib/Elevate/Blockers/Repositories.pm
```diff
@@ -94,6 +94,14 @@ sub _yum_is_stable ($self) {
...
 
+        my $error_msg = <<~'EOS';
+        '/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.
+
+        You may want to consider reaching out to cPanel Support for assistance:
```
sbillik, line 100: See previous comments.

t/blocker-Repositories.t
```diff
@@ -205,18 +205,26 @@ is $yum->_check_yum_repos() => { $unvetted => 1, $rpms_from_unvetted => 1, $inva
         },
     );
 
+    my $error_msg = <<~'EOS';
+    '/usr/bin/yum makecache' failed to return cleanly.  This could be due to a temporary mirror problem or it could be indicative of a larger issue such as a broken repository.  Since this script relies heavily on yum, this will need to be addressed before upgrading.
+
+    You may want to consider reaching out to cPanel Support for assistance:
```
sbillik, line 211: See previous comments.

elevate-cpanel
davelcpanelnet: Done

elevate-cpanel
davelcpanelnet: Done

t/blocker-Repositories.t
davelcpanelnet: Done

cPholloway (approved):

cPholloway (changes requested):
It looks like you put the DOCS changes directly into `elevate-cpanel` instead of putting them in `Elevate::Blockers::Repositories` and running make ot update the main script.

NOTE:  You should never directly modify the main `elevate-cpanel` script

cPholloway (approved):

annes449 (approved):
QA is all good.

PR #455, opened 2024-06-17
Remove python36 blocker

sfbee:
Case RE-336: Remove the python36 blocker, as it has been fixed upstream.

Changelog: Removed python36 blocker on CloudLinux/CentOS 7


lib/Elevate/OS/CloudLinux7.pm
```diff
@@ -22,7 +22,6 @@ use constant leapp_repo_beta                 => 'elevate-updates-testing';
 use constant leapp_can_handle_epel           => 1;
 use constant leapp_can_handle_imunify        => 1;
 use constant leapp_can_handle_kernelcare     => 1;
-use constant leapp_can_handle_python36       => 1;
```
cPholloway, line 25: FIX:  Let's go ahead and kill this key off from `Elevate::OS` completely.

There are two more spots that it needs to be removed from:
```
root@a8 ~/git/elevate (RE-336) # git grep 'leapp_can_handle_python36' lib/ script/
lib/Elevate/OS.pm:        'leapp_can_handle_python36',          # This is used to determine if we can skip the python36 blocker or not
lib/Elevate/OS/RHEL.pm:use constant leapp_can_handle_python36       => 0;
```

cPholloway (changes requested):

cPholloway (approved):

toddr (approved):

annes449 (approved):
All good with QA

PR #456, opened 2024-06-18
Use 'CloudLinux_8' target to backup EA4 profile with I360 installed

cPholloway:
Case RE-403:  In RE-313, we added support for Imunify 360's hardened PHP repo to the EA4 blocker.  Unfortunately, this improvement did not account for some packages that Imunify 360's hardened PHP repo does not provide which has resulted in some servers making it past the EA4 blocker that had packages installed that were not supported on A8/CL8. This change removes the current logic to support Imunify 360's hardened PHP repo, and replaces it by teaching
'Elevate::EA4::_backup_ea4_profile' to use the 'CloudLinux_8' target when Imunify 360 is installed with the hardened PHP feature.  We can do this since Imunify 360's hardened PHP feature is essentially just providing CloudLinux's EA4 to servers with the feature available.

Changelog: Use 'CloudLinux_8' target to backup EA4 profile on servers
 with Imunify 360 installed and providing the hardened PHP feature

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1064,7 +1064,6 @@ EOS
         return unless scalar keys $dropped_pkgs->%*;
 
         my @incompatible;
-        my @imunify_pkgs;
         foreach my $pkg ( sort keys %$dropped_pkgs ) {
             my $type = $dropped_pkgs->{$pkg} // '';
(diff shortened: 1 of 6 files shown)


toddr (approved):

annes449 (approved):
QA is all good!

PR #457, opened 2024-06-20
Teach wait_for_leapp_completion() about the no-leapp option

cPholloway:
Case RE-475:  In case RE-224, we added a check to ensure that leapp has completed before continuing with stage 4 / starting the post leapp components.  Unfortunately, this check did not account for the no-leapp option that the script provides for those who want to perform the upgrade from C7->C8 without using the leapp utility, but still want the script to handle the cPanel specific potions of the upgrade.  This resulted in upgrades using the no-leapp option to fail in stage 4 while waiting on leapp to complete.  This change teaches this check about the no-leapp option which allows for upgrades using this option to skip past this check.

Changelog: Teach wait_for_leapp_completion() about the no-leapp option

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -6081,6 +6081,9 @@ EOS
     }
 
     sub wait_for_leapp_completion ($self) {
+
+        return 1 unless $self->cpev->should_run_leapp();
+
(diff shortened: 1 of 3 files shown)


toddr (approved):

annes449 (approved):
All good with QA

PR #461, opened 2024-06-27
Do not allow check to run in the middle of an elevation

cPholloway:
Case RE-505:  Previously, a user or security advisor could execute --check in the middle of a failed elevation.  It does not make sense for this to be allowed since the elevation attempt has already made it past the blocker stage.  At this point, it is far more appropriate for the user to execute --continue in order to complete the elevation.

Of note, one of the side affects of this is that --check will clear the Elevate::OS cache which will cause failures if the elevation attempt has already made it to stage 4 or 5 and the OS has been upgraded, but cPanel services and friends have not been restored.

Changelog:  Do not allow check to run in the middle of an elevation

### DIFF
--- a/lib/Elevate/Blockers.pm
+++ b/lib/Elevate/Blockers.pm
@@ -90,6 +90,17 @@ sub check ($self) {    # do_check - main  entry point
         return 1;
     }
 
+    my $stage = Elevate::Stages::get_stage();
+    if ( $stage != 0 && $stage <= cpev::VALID_STAGES() ) {
+        die <<~"EOS";
+        An elevation process is currently in progress: running stage $stage
+        You can check the log by running:
+            /scripts/elevate-cpanel --log
+        or check the elevation status:
+            /scripts/elevate-cpanel --status
+        EOS
+    }
+
     Elevate::Blockers::Distros::bail_out_on_inappropriate_distro();
 
     # If no argument passed to --check, use default path:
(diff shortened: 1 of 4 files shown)


lib/Elevate/Blockers.pm
```diff
@@ -90,6 +90,17 @@ sub check ($self) {    # do_check - main  entry point
         return 1;
     }
 
+    my $stage = Elevate::Stages::get_stage();
```
toddr, line 93: I guess I don't understand why `$self->cpev->service->is_active` isn't enough?

toddr (approved):

lib/Elevate/Blockers.pm
cPholloway: Honestly, I did not dig further into why `$self->cpev->service->is_active` wasn't enough, but there very well could be a bug in that routine.

I went with the fix that I did because I already had my head wrapped around this fix before I started, but I could go back investigate why `is_active` is not catching this issue if you would like?

toddr: I suspect it doesn't catch before the service is in place, but the bug we are fixing definitely happened when the service was present...

cPholloway: Yup, my guess is that routine has a bug in it that is preventing it from catching while the service is in place.  This change definitely resolves the issue, but If you think it is worth the time to debug why `is_active` is not catching the issue, then I can replicate the issue again (it takes a bit to get a server into this state), and debug that routine to see why it is not enough.

Let me know if you think it is worth the time.

annes449 (approved):
QA is all good.

PR #462, opened 2024-07-03
Block when packages are installed from a disabled repo

cPholloway:
Case RE-420:  Previously, we ignored packages that were installed directly via the RPM binary or packages that were installed from repositories that had since been disabled.  This led to elevations being allowed to continue when they should have been blocked due to packages being installed from unvetted sources.  This could lead to services or other things being broken on the server after the elevation has completed.  This change makes it so that we detect when packages are installed from disabled repositories and when packages are installed directly via the RPM binary.

Changelog:  Block when packages are installed from a disabled repo

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -133,6 +133,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
 
     use Simple::Accessor qw(
       blockers
+      yum
     );
 
     # use Log::Log4perl qw(:easy);
@@ -147,6 +148,10 @@ BEGIN {    # Suppress load of all of these at earliest point.
         Carp::confess(q[Missing blockers]);
     }
 
+    sub _build_yum ($self) {
+        return Elevate::YUM->new( cpev => $self );
+    }
+
     sub cpev ($self) {
         return $self->blockers->cpev;
     }
@@ -158,6 +163,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
           should_run_leapp
           ssystem
           ssystem_capture_output
+          ssystem_hide_and_capture_output
         };
 
         foreach my $subname (@_DELEGATE_TO_CPEV) {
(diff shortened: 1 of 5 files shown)


elevate-cpanel
```diff
@@ -7572,6 +7595,50 @@ EOS
...
+
+        my @extra_packages;
+        while ( my $line = shift @lines ) {
+            chomp $line;
+            my ( $package, $version, $repo ) = split( qr{\s+}, $line );
+
+            if ( !length $version ) {
+                my $extra_line = shift @lines;
+                chomp $extra_line;
+                $extra_line =~ s/^\s+//;
```
troglodyne, line 7615: Interesting that we'd even get word wrap here. Only makes sense in the context that we unshift `/usr/bin/bash -c` to our arglist in `_ssystem`, which arguably should be optional/overridable to avoid the need for shenanigans like this.

Or... since we are running via bash, you could always run `stty cols 256 && cmd` or something similarly large to avoid wrap ever happening maybe?

elevate-cpanel
```diff
@@ -7572,6 +7595,50 @@ EOS
...
+                $repo =~ s/\s+//g;
+            }
+            length $repo or next;    # We screwed up the parse. move on.
+
+            $repo eq 'installed' or next;
+
+            $package =~ s/\.(noarch|x86_64)$//;
+            my $arch = $1 // '?';
+
+            next if $package =~ m/^cpanel-/;
```
troglodyne, line 7630: the golfer in me cries out for either a `|` in that regexp or use of `grep` :P

t/blocker-Repositories.t
```diff
@@ -319,4 +319,76 @@ is $yum->_check_yum_repos() => { $unvetted => 1, $rpms_from_unvetted => 1, $inva
...
 
+{
+    note 'Testing pacakges installed without associated repos blocker';
+
+    my $stdout = <<'EOS';
+Loaded plugins: fastestmirror, universal-hooks
+Loading mirror speeds from cached hostfile
+ * cpanel-addons-production-feed: 184.94.196.92
+ * epel: mirror.compevo.com
+Extra Packages
```
troglodyne, line 330: Should probably have a line with wrap on all three given you accounted for all three wrapping?

troglodyne (approved):
Seems like it should be fine, though I'd have preferred if we could have avoided the more complex line parsing somehow

annes449 (approved):
Qa is all good.

PR #465, opened 2024-07-05
Use croak() instead of die() in ssystem_and_die

sloanebernstein:
Give better context when `ssystem_and_die` blows up as requested.

Additionally, update the tags file generation target in the Makefile to handle the now-modular nature of the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ cover:
 	find cover_db -type d -exec chmod 755 {} \;
 
 tags:
-	/usr/bin/ctags -R --languages=perl elevate-cpanel t
+	/usr/bin/ctags -R --languages=perl --extra=+q script lib t
 
 elevate-cpanel: $(wildcard lib/**/*) script/elevate-cpanel.PL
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ cover:
 	find cover_db -type d -exec chmod 755 {} \;
 
 tags:
-	/usr/bin/ctags -R --languages=perl elevate-cpanel t
+	/usr/bin/ctags -R --languages=perl --extra=+q script lib t
```
toddr, line 31: what specifically is broken here or being fixed?

sloanebernstein: What's being fixed here is my (and others') ability to use Ctrl-] / Ctrl-T in vim to jump to definitions. It's totally unrelated but also absolutely immaterial to the user-facing functionality or the build/release process. If you really want a separate PR for this...

cPholloway (approved):

toddr (approved):

cPholloway:
QA is good

PR #473, opened 2024-07-11
Have ELS component always remove the ELS repo files if they exist

cPholloway:
Case RE-538:  Previously, we would check to see if the 'els-define' package was installed and return early in this pre_leapp component. However, due to the CentOS 7 base repo moving to vault, it was found that ELS was installed while yum was in a broken state which resulted in the 'els-define' package not getting installed.  As such, this change makes it so that we always remove the ELS repo files if they exist since we can not rely on the 'els-define' package being installed.

Changelog: Have ELS component always remove the ELS repo files if they
 exist

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -5441,8 +5441,6 @@ EOS
 
         return unless Elevate::OS::remove_els();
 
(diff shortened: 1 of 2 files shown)


toddr (approved):

annes449 (approved):
QA is all good.

PR #479, opened 2024-07-19
Check whether leapp can control the boot process

sloanebernstein:
Case RE-514: Currently, the script assumes that leapp will have sufficient control over the boot process via GRUB2 configuration. This can result in upgrades making it all the way to leapp only to discover that leapp cannot boot into the custom environment leapp uses to carry out the upgrade, leading to a system that largely cannot be restored into a fully functional state.

It is possible to check that changes made to the bootloader are reflected in the behavior of the boot process, but this requires doing work prior to any major changes to the system. Stage 1 is an ideal time for this, but also stage 1 was not designed to result in a reboot, unlike every other stage. To make this work, the bootstraping of the elevate-cpanel service has been moved to a pre-stage, and the start and continuation entry points have been changed not to call stage 1 directly. This allows the stage 1 subroutine to return a signal to reboot like all the other stages, though it also results in some cosmetic and (hopefully) minor changes to user-facing behavior, some of which can be mitigated.

In particular, the check being done is to use the `grubby` utility to add and later remove a kernel argument in the default boot entry. The argument is designed to be exceedingly unlikely to ever be something that the kernel or PID 1 would recognize, and thus they ignore it; the script then checks for the argument via `/proc/cmdline` at the very start of stage 2. If found, the rest of stage 2 proceeds as normal; if not, the entire ELevate process is reset as if `--start` had never been passed to the script, and a custom failure notification is emitted.

Changelog: Check whether leapp can control the boot process before
 significant and irreversable changes are applied to the system.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -3524,6 +3524,7 @@ EOS
     use Cpanel::JSON            ();
     use Cpanel::Pkgr            ();
     use Cpanel::SafeRun::Simple ();
+    use Cpanel::SafeRun::Object ();
 
     # use Elevate::Components::Base();
     our @ISA;
@@ -3533,6 +3534,118 @@ EOS
 
     use constant GRUB_EDITENV  => '/usr/bin/grub2-editenv';
     use constant GRUB_ENV_FILE => '/boot/grub2/grubenv';
+
+    use constant GRUBBY_PATH  => '/usr/sbin/grubby';
+    use constant CMDLINE_PATH => '/proc/cmdline';
+
+    use constant EX_UNAVAILABLE => 69;
+
+    sub _call_grubby ( $self, @args ) {
+
+        my %opts = (
+            should_capture_output => 0,
+            should_hide_output    => 0,
+            die_on_error          => 0,
+        );
+
+        if ( ref $args[0] eq 'HASH' ) {
+            my %opt_args = %{ shift @args };
+            foreach my $key ( keys %opt_args ) {
+                $opts{$key} = $opt_args{$key};
+            }
+        }
+
+        unshift @args, GRUBBY_PATH;
+
+        return $opts{die_on_error} ? $self->ssystem_and_die(@args) : $self->ssystem( \%opts, @args );
+    }
+
+    sub _default_kernel ($self) {
+        return $self->_call_grubby( { should_capture_output => 1, should_hide_output => 1 }, '--default-kernel' )->{'stdout'}->[0] // '';
+    }
+
+    sub _persistent_id {
+        my $id = Elevate::StageFile::read_stage_file( 'bootloader_random_tag', '' );
+        return $id if $id;
+
+        $id = int( rand(100000) );
+        Elevate::StageFile::update_stage_file( { 'bootloader_random_tag', $id } );
+        return $id;
+    }
+
+    sub mark_cmdline ($self) {
+        my $arg = "elevate-" . _persistent_id;
+        INFO("Marking default boot entry with additional parameter \"$arg\".");
+
+        my $kernel_path = $self->_default_kernel;
+        $self->_call_grubby( { die_on_error => 1 }, "--update-kernel=$kernel_path", "--args=$arg" );
+
+        return;
+    }
+
+    sub _remove_but_dont_stop_service ($self) {
+
+        $self->cpev->service->disable();
+        $self->ssystem( '/usr/bin/systemctl', 'daemon-reload' );
+
+        return;
+    }
+
+    sub verify_cmdline ($self) {
+        if ( $self->cpev->should_run_leapp() ) {
+            my $arg = "elevate-" . _persistent_id;
+            INFO("Checking for \"$arg\" in booted kernel's command line...");
+
+            my $kernel_cmdline = eval { File::Slurper::read_binary(CMDLINE_PATH) } // '';
+            DEBUG( CMDLINE_PATH . " contains: $kernel_cmdline" );
+
+            my $detected = scalar grep { $_ eq $arg } split( ' ', $kernel_cmdline );
+            if ($detected) {
+                INFO("Parameter detected; restoring entry to original state.");
+            }
+            else {
+                ERROR("Parameter not detected. Attempt to upgrade is being aborted.");
+            }
+
+            my $kernel_path = $self->_default_kernel;
+            my $result      = $self->_call_grubby( "--update-kernel=$kernel_path", "--remove-args=$arg" );
+            WARN("Unable to restore original command line. This should not cause problems but is unusual.") if $result != 0;
+
+            if ( !$detected ) {
+
+                my $stage              = Elevate::Stages::get_stage();
+                my $pretty_distro_name = $self->cpev->upgrade_to_pretty_name();
+                my $msg                = <<"EOS";
+The elevation process failed during stage $stage.
+
+Specifically, the script could not prove that the system has control over its
+own boot process using the utilities the operating system provides.
+
+For this reason, the elevation process has terminated before making any
+irreversible changes.
+
+You can check the error log by running:
+
+    $0
+
+Before you can run the elevation process, you must provide for the ability for
+the system to manipulate its own boot process. Then you can start the elevation
+process anew:
+
+    $0 --start
+
+EOS
+                Elevate::Notify::send_notification( qq[Failed to update to $pretty_distro_name] => $msg );
+
+                $self->cpev->do_cleanup(1);
+                $self->_remove_but_dont_stop_service();
+
+                exit EX_UNAVAILABLE;    ## no critic(Cpanel::NoExitsFromSubroutines)
+            }
+        }
+
+        return;
+    }
 
     sub pre_leapp ($self) {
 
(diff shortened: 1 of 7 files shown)


troglodyne (approved):
I really should have waited till I drank coffee this morning on this...

Nothing obviously wrong but I think I confused myself when reading over this a few times.

Overall considering the scope of the change I'd normally say "I'm surprised this did not cause tests to fail," but I know well the state of the t/ dir. This *is* however the perfect time to add tests for the updated/added functionality. I leave it to your discretion as to whether you have time for that or not, but *strongly* recommend it.

elevate-cpanel
```diff
@@ -8429,37 +8528,40 @@ sub run_service_and_notify ($self) {    # FIXME: move to Elevate::Service - need
...
-        Elevate::StageFile::update_stage_file( { status => q[failed] } );
-        $self->_notify_error($error);
+            Elevate::StageFile::update_stage_file( { status => q[failed] } );
+            $self->_notify_error($error);
 
-        LOGDIE($error);
-    }
+            LOGDIE($error);
+        }
+    } while ( $action_todo == ACTION_CONTINUE );
```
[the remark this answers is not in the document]
sloanebernstein: I wanted to emphasize the aspect of "this block must run at least once" which I think is better communicated by do-while.

elevate-cpanel
```diff
@@ -3534,6 +3534,106 @@ EOS
...
+    $0
+
+If it is possible to restore the ability for the system to manipulate the boot process, you can restart the elevation process anew:
+
+    $0 --start
+
+EOS
+            Elevate::Notify::send_notification( qq[Failed to update to $pretty_distro_name] => $msg );
+
+            $self->cpev->do_cleanup(1);
```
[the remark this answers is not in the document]
sloanebernstein: The issue is that invoking `remove` necessarily invokes `stop`: https://github.com/cpanel/elevate/blob/d312cdc7b88ee46c3ec8965698847e55e8d65039/lib/Elevate/SystemctlService.pm#L76

Since we are invoking `do_cleanup` in the service context, allowing `stop` to run is likely to result in terminating the script before it finishes its cleanup. By contrast, `disable` also removes the service file, but it does not stop the service by default. We can thus let the service terminate on its own by reaching `exit`.

Yes, this is totally brain-dead, but re-doing the service handling code is more scope creep than I needed.

elevate-cpanel
```diff
@@ -3534,6 +3534,106 @@ EOS
...
+
+    sub verify_cmdline ($self) {
+        my $arg = "elevate-" . _persistent_id;
+        INFO("Checking for \"$arg\" in booted kernel's command line...");
+
+        my $kernel_cmdline = eval { File::Slurper::read_binary(CMDLINE_PATH) };
+        LOGDIE( CMDLINE_PATH . ' is unexpectedly empty; leapp may not support this system' ) unless $kernel_cmdline;
+        DEBUG( CMDLINE_PATH . " contains: $kernel_cmdline" );
+
+        my $detected = scalar grep { $_ eq $arg } split( ' ', $kernel_cmdline );
```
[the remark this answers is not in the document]
sloanebernstein: With `grep`, I like being explicit about when I'm using it in scalar context.

sloanebernstein:
> Overall considering the scope of the change I'd normally say "I'm surprised this did not cause tests to fail," but I know well the state of the t/ dir. This _is_ however the perfect time to add tests for the updated/added functionality. I leave it to your discretion as to whether you have time for that or not, but _strongly_ recommend it.

There are basically no tests for the high-level logic and control flow of the script, and it's not exactly written in a way that is conducive to making that happen. Tests for the Grub2 component should be reasonable, though.

elevate-cpanel
```diff
@@ -3534,6 +3534,106 @@ EOS
...
+
+        my $kernel_path = $self->_default_kernel;
+        $self->_call_grubby( { die_on_error => 1 }, "--update-kernel=$kernel_path", "--args=$arg" );
+
+        return;
+    }
+
+    sub verify_cmdline ($self) {
+        my $arg = "elevate-" . _persistent_id;
+        INFO("Checking for \"$arg\" in booted kernel's command line...");
```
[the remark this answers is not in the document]
davelcpanelnet: Seems like backticks have historically the way to add emphasis info INFO() blocks.

script/elevate-cpanel.PL
```diff
@@ -928,20 +930,18 @@ responsible of the multiple reboots.
...
+    Elevate::Motd->setup();
 
-    return 1 unless _sanity_check();
+    if ( !$self->should_run_leapp() ) {
+        Elevate::Stages::bump_stage();
+        return ACTION_CONTINUE;
+    }
 
-    print_box( "Starting stage 1 of " . VALID_STAGES . ": Installing " . $self->service->name . " service" );
+    $self->run_component_once( 'Grub2' => 'mark_cmdline' );
```
cPholloway, line 942: FIX:  We should skip this if they provide the no-leapp option since it has the potential to die

script/elevate-cpanel.PL
```diff
@@ -952,9 +952,7 @@ Update the current distro packages then reboot.
 
 sub run_stage_2 ($self) {
 
-    Elevate::Marker::startup();
-
-    Elevate::Motd->setup();
+    $self->run_component_once( 'Grub2' => 'verify_cmdline' ) if $self->should_run_leapp();
```
cPholloway, line 955: TRIVIAL/PREFERENCE:  I think it would make the code easier to read if we executed this component no matter what.  Then, just have the first line of `verify_cmdline` be `return unless $self->should_run_leapp();`

lib/Elevate/Components/Grub2.pm
cPholloway, line 1: FIX:  Please add unit tests for the changes to this file.  Honestly, MockFile for the contents of `/proc/cmdline` is going to be the best testing that we are likely to get with this.

cPholloway (changes requested):

lib/Elevate/Components/Grub2.pm
```diff
@@ -29,6 +29,110 @@ use Elevate::Blockers ();
...
+
+You can check the error log by running:
+
+    $0
+
+If it is possible to restore the ability for the system to manipulate the boot process, you can restart the elevation process anew:
+
+    $0 --start
+
+EOS
```
cPholloway, line 123: @WriterAmanda - please review this string

script/elevate-cpanel.PL
sloanebernstein: Already handled in the block above this by `should_run_leapp`.

cPholloway: ah, missed that.

lib/Elevate/Components/Grub2.pm
```diff
@@ -29,6 +29,110 @@ use Elevate::Blockers ();
...
+
+        # Can't use _notify_error as in run_service_and_notify, because that
+        # tells you to use --continue, which won't work here due to the
+        # do_cleanup invocation:
+        my $stage              = Elevate::Stages::get_stage();
+        my $pretty_distro_name = $self->cpev->upgrade_to_pretty_name();
+        my $msg                = <<"EOS";
+The elevation process failed during stage $stage.
+
+Specifically, it was not possible to prove that the system has control over its own boot process using the utilities the operating system provides.
```
WriterAmanda, line 111:
```suggestion
Specifically, it could not prove that the system has control over its own boot process using the utilities the operating system provides.
```

lib/Elevate/Components/Grub2.pm
```diff
@@ -29,6 +29,110 @@ use Elevate::Blockers ();
...
+
+For this reason, the elevation process has been terminated before making any unreversable changes.
```
WriterAmanda, line 113:
```suggestion
For this reason, the elevation process terminated before making any unreversable changes.
```

lib/Elevate/Components/Grub2.pm
```diff
@@ -29,6 +29,110 @@ use Elevate::Blockers ();
...
+
+You can check the error log by running:
+
+    $0
+
+If it is possible to restore the ability for the system to manipulate the boot process, you can restart the elevation process anew:
```
WriterAmanda, line 119:
```suggestion
Before you can run the elevation process, you must restore the ability for the system to manipulate the boot process. Then you can restart the elevation process anew: 
```

WriterAmanda (changes requested):
I've added suggested changes to eliminate passive voice where possible and improve clarity.

lib/Elevate/Components/Grub2.pm
toddr: this is somewhat of a change in meaning... :/

WriterAmanda: Feel free to ignore any of my suggestions that change the meaning. The goal is still technical accuracy :)

troglodyne:
Added test looks good to me, probably time to take another look @cPholloway

cPholloway (approved):
Test looks good.  Thank you for adding this 👍

PR #481, opened 2024-07-25
Do not log output of Imunify license check

cPholloway:
Case RE-404: The output of the Imunify license check could be really noisy in the output while in check mode making it more difficult to pick out if a blocker has been found.  This change makes it so that the output of this command is no longer printed to STDOUT or logged.

Changelog: Do not log output of Imunify license check

### DIFF
--- a/t/blocker-Imunify.t
+++ b/t/blocker-Imunify.t
@@ -36,7 +36,7 @@ my $imunify = $cpev->get_blocker('Imunify');
     my @cmds;
     my @stdout;
     $cpev_mock->redefine(
-        ssystem_capture_output => sub ( $, @args ) {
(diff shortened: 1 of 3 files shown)


t/blocker-Imunify.t
```diff
@@ -36,7 +36,7 @@ my $imunify = $cpev->get_blocker('Imunify');
     my @cmds;
     my @stdout;
     $cpev_mock->redefine(
-        ssystem_capture_output => sub ( $, @args ) {
+        ssystem_hide_and_capture_output => sub ( $, @args ) {
```
davelcpanelnet, line 39: Doesn't get any simpler : )

timmullin (approved):

PR #489, opened 2024-08-02
Restore EA4 config files based on package prefix

### LINKED ISSUE #486: [BUG] PHP timezone is changed from Australia/Sydney (etc) to UTC post-ELevate
**Describe the bug**
ELevate process does not seem to fully restore php.ini values into the system files, hence things like timezone are lost.

I thought that it was taking full backups of the files and then restoring them? Server time is OK, it's just PHP that gets lost.

**To Reproduce**
1. Upgrade server via ELevate.
2. Crons fire at incorrect time.
3. PHP timezone revert to UTC.

**Expected behavior**
php.ini backups for all versions being restored should bring the customisations back, and avoid impact to applications.

cPholloway:
Case RE-499: The restore logic for EA4 config files was broken when the packages were provided by either CloudLinux's EA4 repo or Imunify's hardened PHP repo.  This was because the restore logic was gathering the EA4 packages based on the repoid that they came from, but the regex only took cPanel's version of EA4 into account.  This change makes it so that we gather these packages based on the package prefix of 'ea-' so that the config files for the installed packages are always gathered regardless of which repo provides them.

Fixes #486

Changelog: Fix restore logic for config files of packages provided by
 EA4

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -3441,8 +3441,7 @@ EOS
 
         Elevate::StageFile::remove_from_stage_file('ea4_config_files');
 
-        my $ea4_regex        = qr/^EA4(:?-c7)?/a;
-        my $ea4_config_files = $self->rpm->get_config_files_for_repo($ea4_regex);
+        my $ea4_config_files = $self->rpm->get_config_files_for_pkg_prefix('ea-');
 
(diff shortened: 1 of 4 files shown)


toddr (approved):

troglodyne:
Testing checks out good on my end, I think it should be merged

PR #490, opened 2024-08-05
Move logic to gather and store php usage to a component

cPholloway:
Case RE-611: With the changes to the logic to support Imunify 360 on CentOS 7, it was found that sites using older hardened PHP versions no longer had the correct version of PHP set after elevate completed.  This was because the changes to the Imunify 360 logic made it so that the code path that gathered and stored this information was no longer reachable.  As such, this logic has been refactored to only check for PHP versions that would otherwise be dropped after the upgrade to AlmaLinux 8 that are not in use and the logic to gather and store the current PHP usage has been moved to the EA4 component so that it always gets executed and thus the logic to set the PHP version after the elevation will always get executed now too.

Changelog: Move logic to gather and store php usage to a component

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1114,43 +1114,30 @@ EOS
     }
 
     sub _php_version_is_in_use ( $self, $php ) {
-        my $current_php_usage = $self->_get_php_versions_in_use();
+        my $current_php_usage = $self->_get_php_usage();
 
         return 1 if $current_php_usage->{api_fail};
 
         return $current_php_usage->{$php} ? 1 : 0;
     }
 
-    our $php_versions_in_use;
+    our $php_usage;
 
-    sub _get_php_versions_in_use ($self) {
-        return $php_versions_in_use if defined $php_versions_in_use && ref $php_versions_in_use eq 'HASH';
+    sub _get_php_usage ($self) {
+        return $php_usage if defined $php_usage && ref $php_usage eq 'HASH';
 
-        my $out    = Cpanel::SafeRun::Simple::saferunnoerror(qw{/usr/local/cpanel/bin/whmapi1 --output=json php_get_vhost_versions});
-        my $result = eval { Cpanel::JSON::Load($out); } // {};
-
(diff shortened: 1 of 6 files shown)


elevate-cpanel
```diff
@@ -3516,6 +3504,13 @@ EOS
         return;
     }
 
+    sub _gather_php_usage ($self) {
+        my $php_get_vhost_versions = Elevate::EA4::php_get_vhost_versions();
+        Elevate::StageFile::remove_from_stage_file('php_get_vhost_versions');
+        Elevate::StageFile::update_stage_file( { php_get_vhost_versions => $php_get_vhost_versions } );
```
atoomic, line 3510: do we want to store it if `undef`?

atoomic (approved):

elevate-cpanel
cPholloway: it shouldn't hurt anything to store it if it is `undef` since the post_leapp code that restores it returns unless it gets an aref back from this

elevate-cpanel
```diff
@@ -1114,43 +1114,30 @@ EOS
...
-        Unable to determine if PHP versions that will be dropped are in use by
-        a domain.  Assuming that they are in use and blocking to be safe.
-
-        EOS
-
-            $php_versions_in_use->{api_fail} = 1;
-            return $php_versions_in_use;
+        my $php_get_vhost_versions = Elevate::EA4::php_get_vhost_versions();
+        if ( !defined $php_get_vhost_versions ) {
+            $php_usage->{api_fail} = 1;
```
troglodyne, line 1131: Bah, apparently I'm the stupid one today. This kind of autovivification of the hashref in fact works just fine (works fine for arrayref too, no matter the position you assign, even if that populates the previous entries as undef). Only issue comes if it is *of the wrong type*, which you do catch here already.

troglodyne (approved):

annes449 (approved):
QA is all good.

davelcpanelnet (commented):
Looks good

PR #492, opened 2024-08-13
RE-639: Fix makefile paths.

davelcpanelnet:
- Should now work on any OpenStack image of a cPanel install after v110.
- Now uses cpm for module dependency installs.
- All Path hardcoding has been removed and replaced with Globals where appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 GIT ?= /usr/local/cpanel/3rdparty/bin/git
 RELEASE_TAG ?= release
 PERL_BIN=/usr/local/cpanel/3rdparty/perl/536/bin
+PERL_LIB=/usr/local/cpanel/3rdparty/perl/536/lib
 VERSION=`cat version`
 
 all:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,17 +13,28 @@ all:
...
-	/usr/local/cpanel/3rdparty/bin/prove t/00_load.t
-	/usr/local/cpanel/3rdparty/bin/yath test -j8 t/*.t
+	$(PERL_BIN)/prove t/00_load.t
+	$(PERL_BIN)/yath test -j8 t/*.t
+
+build:
+    curl -fsSL https://raw.githubusercontent.com/skaji/cpm/main/cpm > /usr/sbin/cpm
+	chmod -v +x /usr/sbin/cpm
+	/scripts/update_local_rpm_versions --edit target_settings.perl-enhanced installed
+	cpm install --with-all --cpanfile t/cpanfile
```
toddr, line 28: I think you're not specifying perl and going to end up with installing to system perl.

Makefile
```diff
@@ -12,17 +13,28 @@ all:
...
 
 test: sanity
 	-$(MAKE) elevate-cpanel
-	/usr/local/cpanel/3rdparty/bin/prove t/00_load.t
-	/usr/local/cpanel/3rdparty/bin/yath test -j8 t/*.t
+	$(PERL_BIN)/prove t/00_load.t
+	$(PERL_BIN)/yath test -j8 t/*.t
+
+build:
+    curl -fsSL https://raw.githubusercontent.com/skaji/cpm/main/cpm > /usr/sbin/cpm
```
toddr, line 25: might be better off copying into the repo. it doesn't need to be in sbin.

Makefile
```diff
@@ -12,17 +13,28 @@ all:
...
 	done
 
 test: sanity
 	-$(MAKE) elevate-cpanel
-	/usr/local/cpanel/3rdparty/bin/prove t/00_load.t
-	/usr/local/cpanel/3rdparty/bin/yath test -j8 t/*.t
+	$(PERL_BIN)/prove t/00_load.t
+	$(PERL_BIN)/yath test -j8 t/*.t
+
+build:
```
toddr, line 24: could we have a comment to explain what this target is used for?

Makefile
```diff
@@ -40,15 +52,11 @@ elevate-cpanel: $(wildcard lib/**/*) script/elevate-cpanel.PL
 				       --leave-broken \
 				       script/$@.PL
 	mv script/$@.PL.static $@
-	MARKER="`cat maint/marker`" perl -pi -e 's|^(#!/usr/local/cpanel/3rdparty/bin/perl)|$$1\n\n$$ENV{MARKER}\n|' $@
-	VERSION=${VERSION} perl -pi -e 's/(^use constant VERSION =>) 1;/$$1 $$ENV{VERSION};/' $@
-	perltidy -b -bext="/" $@
+	MARKER="`cat maint/marker`" $(PERL_BIN)/perl -pi -e 's|^(#!/usr/local/cpanel/3rdparty/bin/perl)|$$1\n\n$$ENV{MARKER}\n|' $@
```
toddr, line 55: I'm very unsure if this expands correctly inside the quotes???

Makefile
```diff
@@ -40,15 +52,11 @@ elevate-cpanel: $(wildcard lib/**/*) script/elevate-cpanel.PL
...
+	VERSION=${VERSION} $(PERL_BIN)/perl -pi -e 's/(^use constant VERSION =>) 1;/$$1 $$ENV{VERSION};/' $@
+	$(PERL_BIN)/perltidy -b -bext="/" $@
 	chmod +x $@
-	perl -cw elevate-cpanel
-
-build:
```
toddr, line 49: What was previously using this target?

toddr (changes requested):

Makefile
cPholloway: This target is used to fatpack the script.  I use this very frequently on my sandbox.  Please add another target to perform this task instead of altering the `build` target.

davelcpanelnet: Moved into a target **fatpack**

Makefile
davelcpanelnet: OK added it in and then `make clean` will remove the file.

Makefile
davelcpanelnet: Strange I had the `--global` in a previous commit and managed to yank it.

```
    -g, --global
          install modules into current @INC instead of local/
```

Makefile
davelcpanelnet: That commit line looks awkward - same line is below and just pathed with the PERL_BIN

Makefile
davelcpanelnet: It expands like this:
```
MARKER="`cat maint/marker`" /usr/local/cpanel/3rdparty/perl/536/bin/perl -pi -e 's|^(#!/usr/local/cpanel/3rdparty/bin/perl)|$1\n\n$ENV{MARKER}\n|' elevate-cpanel
```

Makefile
davelcpanelnet: What `cpm` will actually do is default to $PWD/local/lib/perl5/ as where everything is thrown.  

I changed this around as I really only needed to install Test/PerlTidy.pm so I used that $PWD/local/lib/perl5/ path for the copy and then immediately removed $PWD/local/ as that being in place completely obliterates the yath tests.

cPholloway (approved):

PR #494, opened 2024-08-14
RE-618: Automate C7->A8 happy path regression test on PR creation

davelcpanelnet:
- Uses Github Actions and Terraform to create a centos-7 instance running the latest 11.110.0.* build.
- Checks out the source PR branch.
- Copies Elevate from source PR branch into /scripts/elevate-cpanel.
- Updates cPanel License.
- Runs elevate-cpanel.
- Tracks each reboot and cycles through them
- Upon completion of elevate-cpanel prove is run against t/integration/* where only complete.t currently resides and can be heavily expanded with additional tests.
- Upon successful pipeline - the OpenStack instance is destroyed.
- Upon failed pipeline - the OpenStack instance is kept alive for investigation.

### DIFF
--- a/.github/workflows/openstack.yml
+++ b/.github/workflows/openstack.yml
@@ -1,0 +1,330 @@
+on:
+  push:
+    branches:
+      - "main"
+      - '!docs'
+    tags-ignore:
+      - "*"
+  pull_request:
+  workflow_dispatch:
+
+env:
+  TF_VAR_application_credential_id: ${{ secrets.OS_APPLICATION_CREDENTIAL_ID }}
+  TF_VAR_application_credential_secret: ${{ secrets.OS_APPLICATION_CREDENTIAL_SECRET }}
+  TF_VAR_ssh_private_key: ${{ secrets.SSH_PRIVATE_KEY }}
+  TF_VAR_ssh_public_key: ${{ secrets.SSH_PUBLIC_KEY }}
+  TF_VAR_os_auth_region: ${{ secrets.OS_AUTH_REGION }}
+  TF_VAR_os_auth_url: ${{ secrets.OS_AUTH_URL }}
+  TF_VAR_os_interface: ${{ secrets.OS_INTERFACE }}
+  TF_VAR_os_project_domain_name: ${{ secrets.OS_PROJECT_DOMAIN_NAME }}
+  TF_VAR_os_password: ${{ secrets.OS_PASSWORD }}
+  TF_VAR_os_project_id: ${{ secrets.OS_PROJECT_ID }}
+  TF_VAR_os_project_domain_id: ${{ secrets.OS_PROJECT_DOMAIN_ID }}
+  TF_VAR_os_project_name: ${{ secrets.OS_PROJECT_NAME }}
+  TF_VAR_os_region_name: ${{ secrets.OS_REGION_NAME }}
+  TF_VAR_os_username: ${{ secrets.OS_USERNAME }}
+  tf_working_directory: "./.github/workflows/openstack/terraform"
+
+jobs: 
+    terraform_openstack_create:
+        runs-on: self-hosted
+        defaults:
+          run:
+            working-directory: "./.github/workflows/openstack/terraform"
+        outputs:
+            VM_IP: ${{ steps.get_vm_ip.outputs.VM_IP }}
+        steps:
+        - uses: actions/checkout@v4
+        - name: Terraform fmt
+          id: fmt
+          run: |
+            terraform fmt -check
+          continue-on-error: true
+
+        - name: DEBUGVAR
+          id: debugvar
+          run: |
+            env | awk 'tolower($0)~/auth|project_name/'
+
+        - name: Terraform Init
+          id: init
+          run: |
+            terraform init
+
+        - name: Terraform Plan
+          id: plan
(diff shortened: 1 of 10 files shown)


davelcpanelnet:
Pipeline that ran in my fork prior to PR: https://github.com/davelcpanelnet/elevate/actions/runs/10380441467

.github/workflows/openstack.yml
```diff
@@ -0,0 +1,301 @@
...
+          uses: appleboy/ssh-action@v1.0.3
+          with:
+            host: ${{ needs.wait_for_final_reboot.outputs.VM_IP }}
+            username: 'root'
+            key: ${{ secrets.SSH_PRIVATE_KEY }}
+            port: '22'
+            timeout: 5m
+            command_timeout: 1m
+            script: |
+              /usr/local/cpanel/3rdparty/bin/prove -lvm /opt/elevate/t/integration/*
```
cPholloway, line 285: 1. FIX:  This should be:  `/opt/elevate/t/integration/*.t`

2. QUESTION:  Do we think that there will be enough integration tests to warrant folders?  If so, then we may want to use find here to gather the tests:  `/usr/local/cpanel/3rdparty/bin/prove -lvm $(find /opt/elevate/t/integration/ -type f -name '*.t')`

3. CONSIDER:  Should we use `yath` here?

.github/workflows/openstack.yml
cPholloway, line 1: I don't see this running `preconfig` before starting elevate.  We should create a `t/integration/preconfig.pl` file that is a noop now and have it call it so that we can just update the file as needed later on.

.github/workflows/openstack/terraform/run_elevate.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -x
+
+wget -O /scripts/elevate-cpanel https://raw.githubusercontent.com/cpanel/elevate/release/elevate-cpanel
+chmod -v 700 /scripts/elevate-cpanel
+/scripts/elevate-cpanel --check
+/usr/local/cpanel/bin/whmapi1 start_background_mysql_upgrade version=10.6
```
cPholloway, line 8: FIX: This is not needed and should be removed so that we test that MySQL gets updated as expected

.github/workflows/openstack/terraform/run_elevate.sh
```diff
@@ -0,0 +1,14 @@
...
+sleep 120
+perl -pi*.bak -e 's/^CPANEL=.*/CPANEL=11.110/g' /etc/cpupdate.conf
+/scripts/upcp
+sleep 30
+yes | /scripts/elevate-cpanel --start
```
cPholloway, line 13: EDIT: This should be `/scripts/elevate-cpanel --start --skip-cpanel-version-check --non-interactive`

.github/workflows/website.yml
cPholloway, line 1: QUESTION:  Why is this being removed?  This is the CD for our DOCS site and I do not see a replacement for it in this PR

cPholloway (changes requested):

.github/workflows/openstack.yml
davelcpanelnet: I think, for now, t/integration/complete.t was the specific ask with plans to expand.  Expanding then won't be any complication so I think this linear approach works.

Yath def is my preference.

```/usr/local/cpanel/3rdparty/bin/yath test -v -j8 t/integration/*.t```

File: .github/workflows/openstack.yml [hunk 4]
```diff
@@ -0,0 +1,301 @@
+on: [push]
```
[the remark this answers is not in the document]
davelcpanelnet: Correct.  As this sits the pipeline will only run against push events on all branches.  

I went ahead and opened this up to be identical to what testsuite.yml is doing.

I also went ahead and changed the `runs-on` lines from a target of `self-hosted` to `cPanelInternal` to reflect a runner group Todd and I setup.

.github/workflows/openstack/terraform/run_elevate.sh
davelcpanelnet: Fixed this by removing this file as we no longer needed it.  I did change the line elsewhere.

.github/workflows/openstack.yml
davelcpanelnet: Ah as of now there is no need for the preconfig.  As we iterate over this we can go back and easily add that in as a stage when we have more time to flesh out what it exactly would do.

cPholloway: Running preconfig was part of the AC for this case:

```
( Copying this out of the sha directory to /scripts/elevate-cpanel directly - start_elevate job ) run t/preconfig1.sh which will setup the system.
```

cPholloway (approved):
Going to go ahead and approve this without the preconfig change as we can always add that in later and this is definitely a good framework for what we want to do

.github/workflows/openstack.yml
davelcpanelnet: This has been added in.

cPholloway (approved):

cPholloway:
@toddr - AC has been met now, and I verified that https://github.com/cpanel/elevate/pull/497 passes with this change.

I think this is ready to merge now.

PR #495, opened 2024-08-15
Fix text in failure advice when MySQL restore fails

toddr:
Case RE-662

Changelog: Fix text in failure advice when MySQL restore fails


cPholloway (approved):

annes449 (commented):
All good with QA

PR #501, opened 2024-08-22
Restore net-snmp if it had been installed prior to elevate

timmullin:
Case RE-574:

During elevate net-snmp will get uninstalled when mysql-libs gets uninstalled. On AlmaLinux 8, net-snmp does not depend on mysql-libs so we can reinstall it after the upgrade.

Changelog: Ensure that system with net-snmp installed will have it
  installed post-elevate.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -65,6 +65,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
     $INC{'Elevate/Components/ELS.pm'}                = 'script/elevate-cpanel.PL.static';
     $INC{'Elevate/Components/R1Soft.pm'}             = 'script/elevate-cpanel.PL.static';
     $INC{'Elevate/Components/Panopta.pm'}            = 'script/elevate-cpanel.PL.static';
+    $INC{'Elevate/Components/PackageRestore.pm'}     = 'script/elevate-cpanel.PL.static';
     $INC{'Elevate/OS.pm'}                            = 'script/elevate-cpanel.PL.static';
     $INC{'Elevate/OS/CentOS7.pm'}                    = 'script/elevate-cpanel.PL.static';
     $INC{'Elevate/OS/CloudLinux7.pm'}                = 'script/elevate-cpanel.PL.static';
@@ -6169,6 +6170,67 @@ EOS
 
 }    # --- END lib/Elevate/Components/Panopta.pm
 
+{    # --- BEGIN lib/Elevate/Components/PackageRestore.pm
+
+    package Elevate::Components::PackageRestore;
+
+    use cPstrict;
+
+    use Elevate::StageFile ();
+
+    use Cpanel::Pkgr ();
+
+    # use Elevate::Components::Base();
+    our @ISA;
+    BEGIN { push @ISA, qw(Elevate::Components::Base); }
+
+    sub _get_packages_to_check () {
+        return qw{
+          net-snmp
+        };
+    }
+
+    sub pre_leapp ($self) {
+
+        my @package_list = _get_packages_to_check();
+        my @installed_packages;
+
+        foreach my $package (@package_list) {
+            if ( Cpanel::Pkgr::is_installed($package) ) {
+                push @installed_packages, $package;
+            }
+        }
+
+        my $config_files = $self->rpm->get_config_files( \@installed_packages );
+
+        Elevate::StageFile::update_stage_file(
+            {
+                'packages_to_restore' => $config_files,
+            }
+        );
+
+        return;
+    }
+
+    sub post_leapp ($self) {
+
+        my $package_info = Elevate::StageFile::read_stage_file('packages_to_restore');
+        return unless defined $package_info and ref $package_info eq 'HASH';
+
+        foreach my $package ( keys %$package_info ) {
+
+            $self->dnf->install($package);
+
+            $self->rpm->restore_config_files( @{ $package_info->{$package} } );
+        }
+
+        return;
+    }
+
+    1;
+
+}    # --- END lib/Elevate/Components/PackageRestore.pm
+
 {    # --- BEGIN lib/Elevate/OS.pm
 
     package Elevate::OS;
@@ -7651,12 +7713,12 @@ EOS
 
         my @installed_rpms = $self->get_installed_rpms(q[%{NAME}\n]);
         my @ea_rpms        = grep { $_ =~ qr/^\Q$prefix\E/ } @installed_rpms;
-        my $config_files   = $self->_get_config_files( \@ea_rpms );
+        my $config_files   = $self->get_config_files( \@ea_rpms );
 
         return $config_files;
     }
 
-    sub _get_config_files ( $self, $pkgs ) {
+    sub get_config_files ( $self, $pkgs ) {
 
         my %config_files;
         foreach my $pkg (@$pkgs) {
@@ -7671,6 +7733,7 @@ EOS
             into place.
             EOS
 
+                $config_files{$pkg} = [];
             }
             else {
 
@@ -8758,6 +8821,7 @@ use Elevate::Components::DatabaseUpgrade    ();
 use Elevate::Components::ELS                ();
 use Elevate::Components::R1Soft             ();
 use Elevate::Components::Panopta            ();
+use Elevate::Components::PackageRestore     ();
 
 # - fatpack OS
 use Elevate::OS              ();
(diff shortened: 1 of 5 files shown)


lib/Elevate/Components/PackageRestore.pm
sloanebernstein, line 1: Given that the problem happens due to the `net-snmp` dependency on `mysql-libs`, would it make more sense for this work to be in the MySQL component?

Alternatively, are there other package manipulations that we could justify moving to here in a future case? I'm OK with a new component with the generic name of `PackageRestore`, but it feels weird for this to be the only one, since we're actually restoring plenty of packages elsewhere, but they're usually in the component that is tied to applicable subject matter.

lib/Elevate/Components/PackageRestore.pm
```diff
@@ -0,0 +1,106 @@
...
+    };
+}
+
+#
+# Set as a function for unit testing
+#
+sub _get_files_to_restore ($package) {
+
+    my $package_files_to_restore = {
+        'net-snmp' => [
```
sloanebernstein, line 46: These aren't the only configurable files. Namely, the package also has `/etc/sysconfig/snmpd` and `/etc/sysconfig/snmptrapd` as ways to affect the arguments the services use to start the daemons. Should not those also be backed up?

lib/Elevate/Components/PackageRestore.pm
```diff
@@ -0,0 +1,106 @@
...
+
+        $self->yum->install($package);
+
+        my $restore_files = _get_files_to_restore($package);
+
+        foreach my $restore_file (@$restore_files) {
+            my $backup_file = $restore_file . '.rpmsave';
+
+            if ( -e $backup_file ) {
+                File::Copy::copy( $backup_file, $restore_file )
```
sloanebernstein, line 97: Does this preserve the permissions? By default, those files are 0600, but at least in a quick test, I'm getting 0644 files with the copy.

sloanebernstein (changes requested):

lib/Elevate/Components/PackageRestore.pm
timmullin: I wasn't aware if the files under sysconfig.  I can add those to the list.

lib/Elevate/Components/PackageRestore.pm
timmullin: The issue is that there is more than one place where `net-snmp` gets removed.  The auto-upgrade from MySQL 5.7 to MariaDB 10.6 will remove it.  However, if I upgrade to MariaDB 10.6 before elevating, it will still get removed, probably by the MySQL component somewhere.  Since there were multiple places where it gets removed, i found it prudent to just to handle it in a separate component where I can check for it at the beginning and restore it at the end.  Plus, I also needed to handle restoration of the config files.
Also, I figured that we may discover that there are other packages that get erroneously removed, so I made the module so we can easily add them as we discover them.

timmullin: The dependency of `net-snmp` on `mysql-libs` is kind of strange.  Even though running `rpm -q --whatrequires mysql-libs` on CentOS 7 will list `net-snmp`; I can install it without `mysql-libs` present and it doesn't try to pull it in or even complain.  On AlmaLinux 8, running `rpm -q --whatrequires mysql-libs` does not show `net-snmp` being dependent on `mysql-libs`.

lib/Elevate/Components/PackageRestore.pm
timmullin: I can add a permission check/set to this.

lib/Elevate/Components/PackageRestore.pm
cPholloway: That's probably because `mysql-community-libs` or some other package is meeting the `mysql-libs` dependency:

```
# rpm -q --provides mysql-community-libs
MySQL-shared(x86-64) = 5.7.44-1.el7
libmysqlclient.so.20()(64bit)
libmysqlclient.so.20(libmysqlclient_20.0)(64bit)
libmysqlclient.so.20(libmysqlclient_20.3)(64bit)
mysql-community-libs = 5.7.44-1.el7
mysql-community-libs(x86-64) = 5.7.44-1.el7
mysql-libs = 5.7.44-1.el7
mysql-libs(x86-64) = 5.7.44-1.el7
```

lib/Elevate/Components/PackageRestore.pm
cPholloway: Rather than hardcoding this, it would be better to use `rpm -qc` to get this list or you could make `Elevate::RPM::_get_config_files` public and just call `$self->rpm->get_config_files( [ 'net-snmp' ] );` to get the list.

lib/Elevate/Components/PackageRestore.pm
cPholloway: Nice catch @sloanebernstein 

https://perldoc.perl.org/File::Copy
```
You may use the syntax use File::Copy "cp" to get at the cp alias for this function. The syntax is exactly the same. The behavior is nearly the same as well: as of version 2.15, cp will preserve the source file's permission bits like the shell utility cp(1) would do with default options, while copy uses the default permissions for the target file (which may depend on the process' umask, file ownership, inherited ACLs, etc.). That is, if the destination file already exists, cp will leave its permissions unchanged; otherwise the permissions are taken from the source file and modified by the umask. If an error occurs in setting permissions, cp will return 0, regardless of whether the file was successfully copied.
```

We should be using `File::Copy::cp()` to perform this task.

@timmullin - Please open a follow up case to audit the rest of the code base for this issue too as I suspect we have this issue elsewhere as well.

lib/Elevate/Components/PackageRestore.pm
```diff
@@ -0,0 +1,106 @@
...
+}
+
+sub post_leapp ($self) {
+
+    my $packages = Elevate::StageFile::read_stage_file('packages_to_restore');
+    return unless defined $packages and ref $packages eq 'ARRAY';
+
+    foreach my $package (@$packages) {
+
+        $self->yum->install($package);
```
cPholloway, line 89: TRIVIAL: This should be using `dnf` since it occurs post upgrade

lib/Elevate/Components/PackageRestore.pm
timmullin: That is a good idea.  I didn't know about `rpm -qc`.  That'd be much better.

lib/Elevate/Components/PackageRestore.pm
```diff
@@ -0,0 +1,106 @@
...
+
+        my $restore_files = _get_files_to_restore($package);
+
+        foreach my $restore_file (@$restore_files) {
```
cPholloway, line 93: Could replace this loop with:

```
$self->rpm->restore_config_files(@$restore_files);
```

lib/Elevate/Components/PackageRestore.pm
timmullin: Interesting. If I've upgrade to MariaDB and do `rpm -q --provides MariaDB-compat`, I see mysql-libs in the output list.

lib/Elevate/Components/PackageRestore.pm
timmullin: Created:  https://webpros.atlassian.net/browse/RE-723

davelcpanelnet:
This is good to go.

elevate-cpanel
```diff
@@ -6166,6 +6167,67 @@ EOS
...
 
+{    # --- BEGIN lib/Elevate/Components/PackageRestore.pm
+
+    package Elevate::Components::PackageRestore;
+
+    use cPstrict;
+
+    use Elevate::StageFile ();
+
+    use Cpanel::Pkgr ();
```
davelcpanelnet, line 6178: Stylistic - I always try and alphabetize use statements. Is there a reason there is a space between the statements here?

davelcpanelnet (approved):

annes449 (approved):
QA is all good.

script/elevate-cpanel.PL
```diff
@@ -961,6 +962,10 @@ sub run_stage_2 ($self) {
     $self->ssystem_and_die(qw{/scripts/update-packages});
     $self->ssystem_and_die(qw{/usr/bin/yum -y update});
 
+    # This needs to happen before the database upgrade because
+    # it will remove the packages we want to detect
+    $self->run_component_once( 'PackageRestore' => 'pre_leapp' );
```
toddr, line 967: Sorry to ask this but will this happen before or after the auto-upgrade of MySQL?

timmullin: Before the auto-upgrade.  The auto-upgrade will cause net-snmp to be removed; so I had to detect the package beforehand.

cPholloway (changes requested):
Why was the test `t/components-PackageRestore.t` completely dropped?

We could still add a test for this component asserting:

1.  That it stores a list of packages with their subsequent config files if the package is installed `pre_leapp`
2. That it would call the sub to install any stored packages `post_leapp`
3. That it would call the sub to restore the config files for any stored packages `post_leapp`

annes449 (approved):
QA is all good.

PR #508, opened 2024-09-06
Genericize `--no-leapp`, et al.

sloanebernstein:
Case RE-695: Don't assume that leapp will be the only distro upgrade process we will be wrapping.

Changelog: Turn `--no-leapp` into a more generic
 `--upgrade-distro-manually` option, but continue to recognize
 `--no-leapp` as an alias for it.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -177,7 +177,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
         my @_DELEGATE_TO_CPEV = qw{
           getopt
           upgrade_to_pretty_name
-          should_run_leapp
+          should_run_distro_upgrade
           ssystem
           ssystem_capture_output
           ssystem_hide_and_capture_output
@@ -491,7 +491,7 @@ BEGIN {    # Suppress load of all of these at earliest point.
 
     sub check ($self) {
 
-        return 1 unless $self->should_run_leapp;    # skip when --no-leapp is provided
+        return 1 unless $self->should_run_distro_upgrade;    # skip when --upgrade-distro-manually is provided
 
         my $ok = 0;
         try {
@@ -1158,7 +1158,7 @@ EOS
 
     sub check ($self) {
 
-        return 1 unless $self->should_run_leapp;    # skip when --no-leapp is provided
+        return 1 unless $self->should_run_distro_upgrade;    # skip when --upgrade-distro-manually is provided
 
         my $ok = 1;
         $ok = 0 unless $self->_blocker_grub2_workaround;
@@ -1393,7 +1393,7 @@ EOS
 
     sub check ($self) {    # $self is a cpev object here
 
-        return 0 unless $self->should_run_leapp;
+        return 0 unless $self->should_run_distro_upgrade;
 
         if ( _is_container_envtype() ) {
             return $self->has_blocker( <<~'EOS');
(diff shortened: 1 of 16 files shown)


cPholloway (approved):
Should probably file a DOCS case for this to review the docs/pod changes, but I'm going to go ahead and approve it

elevate-cpanel
```diff
@@ -9067,17 +9068,17 @@ sub start ($self) {
...
 
-    if ( $self->getopt('no-leapp') ) {
+    if ( $self->getopt('upgrade-distro-manually') || $self->getopt('no-leapp') ) {
         my $touchfile = PAUSE_ELEVATE_TOUCHFILE;
         WARN( <<~"EOS");
-        'leapp' checks and run are disabled.
+        Automated upgrade of the distro has been disabled.
         Please wait for the file '$touchfile' to be created.
-        At that stage the elevation process is paused then you can upgrade from 7 to 8 on your own.
+        At that stage, the ELevate process is paused, and you can upgrade from 7 to 8 on your own.
```
davelcpanelnet, line 9076: Ubuntu versions need represented here.    Maybe we need to check Elevate::OS::instance() or something along those lines and show the appropriate text here.  That or we can simply omit the versions here:

`At that stage, the ELevate process is paused, and you can upgrade to the next distro version on your own.`

elevate-cpanel
```diff
@@ -9067,17 +9068,17 @@ sub start ($self) {
...
 
-        Once the system has been successfully upgraded from 7 to 8 you can then remove that file
-        to let the elevation process continue and update cPanel for the updated distribution.
+        Once the system has been successfully upgraded from 7 to 8, you can then remove that file
```
davelcpanelnet, line 9078: Same as above:

Ubuntu versions need represented here. Maybe we need to check Elevate::OS::instance() or something along those lines and show the appropriate text here. That or we can simply omit the versions here:

At that stage, the ELevate process is paused, and you can upgrade to the next distro version on your own.

lib/Elevate/Usage.pm
```diff
@@ -20,7 +20,7 @@ sub _OPTIONS {
       service start clean continue manual-reboots status log check:s
       skip-cpanel-version-check skip-elevate-version-check
       update version
-      no-leapp
+      upgrade-distro-manually no-leapp
```
davelcpanelnet, line 23: s/ no-leapp$//

script/elevate-cpanel.PL
```diff
@@ -554,17 +555,17 @@ sub start ($self) {
...
 
-    if ( $self->getopt('no-leapp') ) {
+    if ( $self->getopt('upgrade-distro-manually') || $self->getopt('no-leapp') ) {
         my $touchfile = PAUSE_ELEVATE_TOUCHFILE;
         WARN( <<~"EOS");
-        'leapp' checks and run are disabled.
+        Automated upgrade of the distro has been disabled.
         Please wait for the file '$touchfile' to be created.
-        At that stage the elevation process is paused then you can upgrade from 7 to 8 on your own.
+        At that stage, the ELevate process is paused, and you can upgrade from 7 to 8 on your own.
```
davelcpanelnet, line 563: Again version specifics regarding ubuntu.

davelcpanelnet (changes requested):

elevate-cpanel
davelcpanelnet: RE-664 looks like it is slated to make the Libs we could rely on for these changes.

Consider any comment here regarding versions for ubuntu in text as gold-plating we have slated for action later.

PR #509, opened 2024-09-10
Add a whitelist for some legacy packages without a repo

cPholloway:
Case RE-603: In RE-420, we started blocking on packages that do not have an associated repo.  This led to a significant spike in the number of servers that were eligible to be elevated out of the box.  Upon looking at the data returned from the blocker added via RE-420, we found that there were several packages that we could whitelist for this blocker which would reduce the number of servers that were blocked out of the box.  This change creates the whitelist and ignores the packages that were added in it.

Changelog: Add a whitelist for some legacy pacakges without a repo

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -1855,6 +1855,24 @@ EOS
     use constant YUM_COMPLETE_TRANSACTION_BIN => '/usr/sbin/yum-complete-transaction';
     use constant FIX_RPM_SCRIPT               => '/usr/local/cpanel/scripts/find_and_fix_rpm_issues';
 
+    use constant EXPECTED_EXTRA_PACKAGES => (
+        qr/^cpanel-/,
+        qr/^easy-/,
+        qr/^kernel/,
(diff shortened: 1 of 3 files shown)


toddr (approved):

elevate-cpanel
```diff
@@ -1855,6 +1855,24 @@ EOS
...
+        'filesystem',
+        'grub',
+        'grubby',
+        'python35',
+        'python38-opt',
+        'virt-what',
+        'vzdummy-systemd-el7',
+      ),
+      Elevate::Constants::R1SOFT_AGENT_PACKAGES;
+
```
davelcpanelnet, line 1875: Awesome I was contemplating the best way to go about the allowed / disallowed packages.

Not sure if you want to revisit https://webpros.atlassian.net/issues/RE-597?jql=text%20~%20%22awscli%22%20AND%20project%20%3D%20RE in another task to handle those cases since this would allow it so easily.

davelcpanelnet (approved):

PR #510, opened 2024-09-11
RE-750: GitHub workflows are stealing jobs from one another.  Fix the Concurrency settings.

davelcpanelnet:
* reboot_watch refactored to quickly check entire log for regex before it begins tail watching.
* reboot stages refactored starting with the first one which was overloaded to start elevate and then watch for a reboot.  Now we start elevate and move to a seperate job to then watch for the reboot and handle things in a more predictable and stable manner.
* Final stage job refactored to explicitly state and act on stage 5.

### DIFF
--- a/.github/workflows/openstack/reboot_watch
+++ b/.github/workflows/openstack/reboot_watch
@@ -2,29 +2,48 @@
 
 use constant ELEVATE_LOG_PATH => '/var/log/elevate-cpanel.log';
 
-use Env;
-
 use File::Tail;
 use POSIX;
 
-my $RETVAL  = 1;
-my $RETRIES = 0;
+my $RETVAL = 1;
 
-while ( $RETVAL != 0 ) {
-    _check_elevate_log_for_regex( ELEVATE_LOG_PATH, ${REGEX}, $RETRIES );
+# Verify REBOOT_STRING isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+# Jump if it's found.
+
+exit 1 if !length( $ENV{REBOOT_STRING} );
+
+open( my $elevate_log_fh, '<', ELEVATE_LOG_PATH ) or die "## [ERROR][reboot_watch]: Unable to open ELEVATE_LOG_PATH: $!\n";
+
+while ( my $line = readline $elevate_log_fh ) {
+    if ( index( $line, $ENV{REBOOT_STRING} ) >= 0 ) {
+        _success_message();
+        close $elevate_log;
+        exit 0;
+    }
 }
 
-sub _check_elevate_log_for_regex {
-    my ( $filepath, $REGEX, $RETRIES ) = @_;
+close $elevate_log_fh;
 
-    my $time = POSIX::strftime( "%Y-%m-%d %H:%M:%S", localtime );
+while ( $RETVAL != 0 ) {
+    _check_elevate_log_for_REBOOT_STRING( ELEVATE_LOG_PATH, $ENV{REBOOT_STRING} );
+    exit 0;
+}
+
+sub _check_elevate_log_for_REBOOT_STRING {
+    my ( $filepath, $REBOOT_STRING, $RETRIES ) = @_;
 
     $file = File::Tail->new( name => $filepath, maxinterval => 1, adjustafter => 7, interval => 1 );
     while ( defined( $line = $file->read ) ) {
-        if ( grep { /$REGEX/m } $line ) {
-            print "## [$time] [INFO]: SUCCESS: Reboot regex ( $REGEX ) found in /var/log/elevate-cpanel.log  ##\n";
-            exit 0;
+        if ( index( $line, $ENV{REBOOT_STRING} ) >= 0 ) {
+            _success_message();
         }
-        $RETRIES++;
     }
 }
+
+sub _success_message {
+    my $time = POSIX::strftime( "%Y-%m-%d %H:%M:%S", localtime );
+    print "## [$time] [INFO]: SUCCESS: Reboot REBOOT_STRING ( $ENV{REBOOT_STRING} ) found in /var/log/elevate-cpanel.log  ##\n";
+    exit 0;
+}
+
+exit 0;
--- a/.github/workflows/openstack.yml
+++ b/.github/workflows/openstack.yml
@@ -1,3 +1,5 @@
+name: Openstack Test Run
+
 on:
   push:
     branches:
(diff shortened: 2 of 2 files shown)


cPholloway (approved):

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,40 @@ use POSIX;
...
 
+# Verify REGEX isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+
+open LOGFILE, ELEVATE_LOG_PATH;
+my @log = <LOGFILE>;
+close LOGFILE;
+
+# Jump if it's found.
+
+if ( grep { /${REGEX}/m } @log ) {
```
troglodyne, line 21: You sure you don't wanna `/\Q${REGEX}\E/m` that? Might make sense so long as you aren't passing in regexp escapes/patterns but just strings to grep for.

troglodyne (approved):
Seems ok presuming this passes.

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,40 @@ use POSIX;
 my $RETVAL  = 1;
 my $RETRIES = 0;
 
+# Verify REGEX isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+
+open LOGFILE, ELEVATE_LOG_PATH;
```
toddr, line 15: Don't use bareword filehandles please.

.github/workflows/openstack/reboot_watch
davelcpanelnet: Making a change for this one.

.github/workflows/openstack/reboot_watch
davelcpanelnet: Got it.

annes449 (commented):
QA test skipped. Dev tools.

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,43 @@ use POSIX;
 my $RETVAL  = 1;
 my $RETRIES = 0;
 
+# Verify REGEX isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+# Jump if it's found.
+
+open my $elevate_log_fh, ELEVATE_LOG_PATH;
```
cPholloway, line 16: FIX: Open the file for reading only and error check if opening the file fails for whatever reason

```
open ( my $elevate_log_fh, '<', ELEVATE_LOG_PATH ) or die "Unable to open ELEVATE_LOG_PATH: $!\n";
```

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,43 @@ use POSIX;
...
+
+while ( !eof($elevate_log_fh) ) {
```
cPholloway, line 18:
```
while ( my $line = readline $elevate_log_fh ) {
```

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,43 @@ use POSIX;
...
+    readline $elevate_log_fh;
+    if (/${REGEX}/) {
```
cPholloway, line 20: Since we are just searching for a string of text here (the regex does not contain metacharacters, it is more secure to wrap this with \Q and \E.

```
if ( $line =~ m/\Q${REGEX}\E/ ) {
```

cPholloway (changes requested):

.github/workflows/openstack/reboot_watch
cPholloway, line 1: QUESTION:  Where exactly is `REGEX` coming from here?

I do not see it declared anywhere in this file, and it seems to be getting passed in as an ENV variable from the CI.  Should we be doing something like:

```
my $regex = $ENV{REGEX};
die "REGEX environment variable is not set\n" unless defined $regex;
```

or is there some CI magic that I missing here?

.github/workflows/openstack/reboot_watch
davelcpanelnet: I can add in error checking if it seems beneficial.

.github/workflows/openstack/reboot_watch
davelcpanelnet: done

.github/workflows/openstack/reboot_watch
davelcpanelnet: [Here is how REGEX is being called from the pipeline.](https://github.com/cpanel/elevate/blob/56cd675b0fddbd10893582c9668b1ecdaedf43be/.github/workflows/openstack.yml#L199) 

Essentially this is a normal CI sort of convention for a watcher where the watch payload is set over and I kept in-line with also passing RETVAL which can be overridden so that it does not parse the log before going into a tailwatch loop.

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,44 @@ use POSIX;
...
 
+# Verify REGEX isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+# Jump if it's found.
+
+exit 1 unless defined $ENV{REGEX};
+
+open my $elevate_log_fh, ELEVATE_LOG_PATH or die "## [ERROR]: Unable to open ELEVATE_LOG_PATH: $!\n";
+
+while ( my $line = readline $elevate_log_fh ) {
+    if ( $line =~ m/\Q${REGEX}\E/ ) {
```
toddr, line 21: This would be safer to be an index check, NOT a regex.

toddr:
```perl
index($line, $ENV{REGEX}) >= 0
```

.github/workflows/openstack/reboot_watch
```diff
@@ -10,21 +10,44 @@ use POSIX;
 my $RETVAL  = 1;
 my $RETRIES = 0;
 
+# Verify REGEX isn't already in the log before we go into tail mode.  The logodump here is sgnificantly faster than File::Tail.
+# Jump if it's found.
+
+exit 1 unless defined $ENV{REGEX};
```
toddr, line 16: `length($ENV{'REGEX'})`

.github/workflows/openstack/reboot_watch
toddr: Please DO NOT use the Env module. You're even later accessing it directly via $ENV{REGEX}. That's a lot confusing.

davelcpanelnet: Removed.

.github/workflows/openstack/reboot_watch
davelcpanelnet: Implemented

.github/workflows/openstack/reboot_watch
davelcpanelnet: Done

PR #511, opened 2024-09-12
Make 'upgrade_to_pretty_name' an Elevate::OS key

cPholloway:
Case RE-697:  This is a refactor to remove the
'upgrade_to_pretty_name()' sub from the main script and make it an Elevate::OS key instead.  Additionally, this change resolves RE-698 by removing a hard coded value for the OS we are upgrading from the service file that we provide.

Changelog: Make 'upgrade_to_pretty_name' an Elevate::OS key


toddr (approved):

davelcpanelnet (approved):
Looks good.  All tests passing and nothing to add.

PR #512, opened 2024-09-13
RE-786: Quick verbiage and formatting fix missing from from RE-750

davelcpanelnet:
Should only add another 30 seconds or so of runtime with the artifacts moving about the stages.

### DIFF
--- a/.github/workflows/openstack.yml
+++ b/.github/workflows/openstack.yml
@@ -161,10 +161,11 @@ jobs:
             cat ${{ github.run_id }}-vm_ip > VM_IP
             ls -la
 
-        - name: Export    to env
+        - name: Export VM_IP to env
           env:
             VM_IP: ${{ steps.VM_IP.outputs.VM_IP }}
           run: echo "VM_IP is ${{ steps.VM_IP.outputs.VM_IP }}"
+
         - name: Setup for Integration Checks Prior to Running Elevate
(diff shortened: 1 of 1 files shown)


troglodyne:
Commit message does not explain why this greatly improves things. Is it because it changes the timing of things, etc.?

Some other questions too... Do artifacts clean themselves up after or is that a manual process? I may not be 'hip' enough to what the CI is doing here yet to competently review this

troglodyne (approved):
Talked with Lanning about the PR and what exactly using artifacts gets us here. Basically it sounds like the runners don't really understand what each other are doing (it's some kind of bank queue) other than "what step am I supposed to be on given what steps have already executed." As such you need to store the IP somewhere that's accessible to the other runners, as we're not guaranteed to pick up the other steps on the same runner PID. Artifacts solve that IPC problem even though that's not their primary purpose. There's probably a more clever way to do this, but good enough is, especially since the public knowing these ephemeral IPs behind devwall doesn't do an attacker much good even if they were to deduce the IP ranges we use internally from this data.

Uploading logs for failed tests as an artifact so they can be linked at the end of the run is probably a good idea as well, not sure if that should be a follow up PR or done prior to merge here.

Anyways, approved.

annes449 (commented):
QA skipped.

PR #513, opened 2024-09-13
Replace the use of leapp in our naming scheme for elevate

timmullin:
Case RE-696:

Since Ubuntu does not rely on leapp to perform distro upgrades, we should change our function names, variable names, etc. to refer to performing a distro upgrade rather than referring to leapp. (Except, of course, with code that actually does interact directly with leapp.)

Changelog:

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -2651,7 +2651,7 @@ EOS
         return %links;
     }
 
-    sub pre_leapp ($self) {
+    sub pre_distro_upgrade ($self) {
 
         $self->ssystem(qw{/usr/bin/ln -snf usr/local/cpanel/scripts /scripts});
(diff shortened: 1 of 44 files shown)


cPholloway:
This is going to conflict with RE-673 / `7a3cce6518f03da28fca393daf89d97ced4d702b`.  You will likely want to pull in that change and then add the rest of your changes on top of it.

cPholloway (changes requested):
This is going to conflict with RE-673 / `7a3cce6518f03da28fca393daf89d97ced4d702b`.  You will likely want to pull in that change and then add the rest of your changes on top of it.

timmullin:
@cPholloway  I have re-based these changes onto RE-695 so as not to have merge conflicts with it.

PR #514, opened 2024-09-23
RE-699: Swap 'return 42' line to Carp::croak()

davelcpanelnet:
RE-699: die() instead of `return 42` within _ssystem().

### DIFF
--- a/lib/Elevate/Roles/Run.pm
+++ b/lib/Elevate/Roles/Run.pm
@@ -74,7 +74,7 @@ sub _ssystem ( $command, %opts ) {
     my @args = @{ $command // [] };
 
     # Only allow the program to be an executable absolute path
-    return 42 if $args[0] !~ '^/' || !-x $args[0];
+    Carp::croak("Program is not an executable absolute path.") if $args[0] !~ '^/' || !-x $args[0];
 
     INFO( "Running: " . join( " ", @args ) );
     INFO();    # Buffer so they can more easily read the output.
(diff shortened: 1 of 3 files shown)


lib/Elevate/Roles/Run.pm
```diff
@@ -74,7 +74,7 @@ sub _ssystem ( $command, %opts ) {
     my @args = @{ $command // [] };
 
     # Only allow the program to be an executable absolute path
-    return 42 if $args[0] !~ '^/' || !-x $args[0];
+    die "Program is not an executable absolute path." if $args[0] !~ '^/' || !-x $args[0];
```
cPholloway, line 77: Let's use `Carp::confess` so that we get a backtrace.

cPholloway (changes requested):

davelcpanelnet:
Works for me - I took the “extremely literal” approach 😂

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lib/Elevate/Roles/Run.pm
davelcpanelnet: Agreed this initial commit was extremely literal in it's use of die().

cPholloway (approved):

toddr (changes requested):
There aren't changes to elevate-cpanel. You need to update it and add it to the commit.

toddr:
make test?

toddr (approved):

cPholloway (approved):

msalinas90 (approved):

PR #516, opened 2024-09-25
Fix: Unable to elevate system while "cl7h" repository is active

lkabuci:

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -6457,7 +6457,8 @@ EOS
             qr/^cloudlinux-ea4-rollout(?:-[0-9]+)?$/,
             'cl-ea4',
             qr/^cl-mysql(?:-meta)?/,
-            'mysqclient', 'mysql-debuginfo'
+            'mysqclient', 'mysql-debuginfo',
+            'cl7h'
         );
 
         my @repos = $self->SUPER::vetted_yum_repo();


elevate-cpanel
```diff
@@ -6457,7 +6457,8 @@ EOS
             qr/^cloudlinux-ea4-rollout(?:-[0-9]+)?$/,
             'cl-ea4',
             qr/^cl-mysql(?:-meta)?/,
-            'mysqclient', 'mysql-debuginfo'
+            'mysqclient', 'mysql-debuginfo',
```
troglodyne, line 6460: Please edit `lib/Elevate/OS/CloudLinux7.pm` with your change and then run `make`. The file you edited is generated by the makefile based off the contents of `lib/`

Thanks!

troglodyne (changes requested):

elevate-cpanel
```diff
@@ -6457,7 +6457,8 @@ EOS
...
+            'cl7h'
```
cPholloway, line 6461: Does CL leapp handle this repo?  Typically just adding a repo to the vetted list is not acceptable as we would need to add a component to handle upgrading the packages from the repo.

toddr:
What is the source of the cl7h repo? We'd need to review it before we green light it.

elevate-cpanel
cPholloway: Okay, this change is fine since it is mapped in CL leapp.  We'll just need you to make the change in `lib/Elevate/OS/CloudLinux7.pm` and run make to regenerate `elevate-cpanel` since `elevate-cpanel` is purely a generated file based on `lib/scripts`

toddr (changes requested):
updating elevate-cpanel by itself will lead to a revert of your changes on the next update. please update the code in lib/

toddr:
I'm replacing this with #518

PR #519, opened 2024-09-27
Merge blockers into components

cPholloway:
Case RE-673: This is a refactor to merge Blockers into Components so that they share a common base.  It also enforces that each Component provides a check, pre_distro_upgrade, and post_distro_upgrade method (even if they are noops).

Changelog: Merge blockers into components


Makefile
sloanebernstein, line 1: Can we get these Makefile changes split out into a separate commit (same PR is fine)?

lib/Elevate/Components.pm
```diff
@@ -75,11 +98,38 @@ our @BLOCKERS = qw{
   AutoSSL
 };
 
-push @BLOCKERS, 'Leapp';    # This blocker has to run last!
+push @CHECKS, 'Leapp';    # This blocker has to run last!
```
sloanebernstein, line 101: The comment is no longer true. Does it still need to be?

lib/Elevate/Components.pm
```diff
@@ -75,11 +98,38 @@ our @BLOCKERS = qw{
...
+
+# This is where to add new components that
+# have checks that are noops
+our @NOOP_CHECKS = qw{
```
sloanebernstein, line 105: Why "no-op"? Seems like a strange choice of a term.

lib/Elevate/Components/AutoSSL.pm
```diff
@@ -16,9 +26,30 @@ use Cpanel::SSL::Auto ();
...
+    return $self->_check_autossl_provider();
+}
+
+sub _check_autossl_provider ($self) {
+
+    if ( $self->is_using_sectigo() ) {
+        WARN( <<~"EOS" );
+        Elevating with Sectigo as the provider for AutoSSL is not supported.
+        If you proceed with this upgrade, we will switch your system
+        to use the Let's Encrypt�~D� provider.
```
sloanebernstein, line 42: UTF-8 mojibake

Since we pretty much force the `C` locale, maybe replace the screwed up ® or ™ with `(R)` or `(TM)`?

lib/Elevate/Components/BootKernel.pm
```diff
@@ -82,4 +89,12 @@ sub _to_str ($e) {
     return $str;
 }
 
+sub pre_distro_upgrade ($self) {
```
sloanebernstein, line 92: Instead of doing this sort of thing with `pre_distro_upgrade`, `post_distro_upgrade`, and `check` in individual components, perhaps these should be added to `Elevate::Components::Base` as default methods? Alternatively, if we think that there's value in having every subclass which doesn't use one of these explicitly override it with an empty method, we could add them to the base class as abstract methods which `die` if called without being overriden.

Makefile
toddr: I'm perfectly ok with another PR for this stuff. Seems unrelated to the work.

cPholloway: I'm going to split the test that necessitated these changes out to RE-857

lib/Elevate/Components.pm
cPholloway: It does.  I have correct this.

lib/Elevate/Components.pm
cPholloway: Previously, there was a test that enforced that every blocker was included in the `BLOCKERS` array.  As part of this refactor, I changed the name of the array to `CHECKS`, but I did not want to remove the test enforcing that every component is included in this array, so I went with `NOOP_CHECKS` to try to convey that these checks should not actually do anything.  I'm definitely open to suggestion on a better way to do this, or we could just consider dropping the test, but my concern about dropping the test is that we end up creating blockers that never get executed.

As it is, I figured the lesser evil is just to enforce calling subs that simply return.

lib/Elevate/Components/BootKernel.pm
cPholloway: I knew I had gotten too close to this.  Adding this to the base class and getting rid of the overrides.

PR #522, opened 2024-10-03
Ensure Imunify upgrade is not executing before uninstalling it

cPholloway:
Case RE-862: The Imunify products have the ability to auto update themselves which can trigger a lock file being in place that will prevent the Imunify installer(s) from executing.  This change makes it so that we check to see if the lock file exists and wait for the PID to complete before attempting to uninstall Imunify in the pre_distro_upgrade stage.

Changelog: Ensure Imunify upgrade is not executing before attempting to
 uninstall it

### DIFF
--- a/lib/Elevate/Components/Imunify.pm
+++ b/lib/Elevate/Components/Imunify.pm
@@ -232,15 +232,29 @@ sub _remove_imunify_360 ($self) {
         return;
     };
 
-    INFO("Imunify360: Removing $product_type prior to upgrade.");
-    INFO("Imunify360: Product $product_type detected. Uninstalling before upgrade for later restore.");
-
     my $installer_script = _fetch_imunify_installer($product_type) or do {
         WARN("Imunify360: Failed to fetch script for $product_type. Skipping upgrade.");
         return;
     };
+
+    # Imunify can auto-update itself.  If this happens, it will
+    # put a lock file in place that will prevent the uninstall from
+    # succeeding.  Ensure it is not updating before attempting
+    # the uninstall
+    my $imunify_pid = Cpanel::LoadFile::load_if_exists('/var/lock/i360deploy.lck') || Cpanel::LoadFile::load_if_exists('/var/lock/imav-deploy.lck');
+    if ($imunify_pid) {
+        INFO("Waiting for Imunify update to complete before uninstalling it");
+        while ( kill 0, $imunify_pid ) {
+            sleep 1;
+        }
+        INFO("Imunify update has completed");
+    }
+
+    INFO("Imunify360: Removing $product_type prior to upgrade.");
+    INFO("Imunify360: Product $product_type detected. Uninstalling before upgrade for later restore.");
+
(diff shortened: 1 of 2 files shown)


lib/Elevate/Components/Imunify.pm
```diff
@@ -232,15 +232,29 @@ sub _remove_imunify_360 ($self) {
...
+        }
+        INFO("Imunify update has completed");
+    }
+
+    INFO("Imunify360: Removing $product_type prior to upgrade.");
+    INFO("Imunify360: Product $product_type detected. Uninstalling before upgrade for later restore.");
+
     if ( $self->ssystem( '/usr/bin/bash', $installer_script, '--uninstall' ) != 0 ) {
-        WARN("Imunify360: Failed to uninstall $product_type.");
+        LOGDIE("Imunify360: Failed to uninstall $product_type.");
```
toddr, line 257: I'm worried that stopping here leaves the system in a broken state. How often do we think this will happen?

cPholloway: > How often do we think this will happen?

With this change, I do not see any reason for this to happen, so my hope is that it doesn't happen again.  That said, it does rely on calling a script that we don't control so it is possible that something out of our control gets broken.

> I'm worried that stopping here leaves the system in a broken state.

It is going to fail and leave the system in a broken state either way.  If we don't die here, then `leapp upgrade` will die in stage 3 because it does not know how to deal with the packages that imunify provides:

```
* 2024-10-01 20:00:43 (7434) [^[[32mINFO^[[0m]             Transaction check:
* 2024-10-01 20:00:43 (7434) [^[[32mINFO^[[0m]
* 2024-10-01 20:00:43 (7434) [^[[32mINFO^[[0m]              Problem: problem with installed package imunify360-php-i360-8.2.2-1.el7.x86_64
* 2024-10-01 20:00:43 (7434) [^[[32mINFO^[[0m]               - cannot install the best update candidate for package imunify360-php-i360-8.2.2-1.el7.x86_64

[ REDACTED ]

* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m] Last Error:
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m]
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m] The 'leapp upgrade' process failed.
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m]
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m] Please investigate, resolve then re-run the following command to continue the update:
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m]
* 2024-10-01 20:00:44 (7345) [^[[31mERROR^[[0m]     /scripts/elevate-cpanel --continue
```

I decided to make it die here because the system is less broken and it give the user a chance to resolve the issue before continuing.  It is also possible that just running `--continue` will resolve the issue since it will run this code again.

toddr (approved):

davelcpanelnet (approved):

PR #524, opened 2024-10-11
Use mv and cp from File::Copy instead of copy and move

cPholloway:
Case RE-723:  As it turns out, File::Copy provides copy and cp that behave slightly different.  The copy and move subs use the umask provided by the running processes shell.  The cp and mv try to preserve permissions for the file that is being copied or moved.  As such, we should prefer cp and mv over copy and move from File::Copy.

Changelog: Use mv and cp from File::Copy instead of copy and move

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -572,8 +572,8 @@ BEGIN {    # Suppress load of all of these at earliest point.
             while ( -e "$link-$rand_uid" && $tries++ < 10000 ) {
                 $rand_uid = Cpanel::UUID::random_uuid();
             }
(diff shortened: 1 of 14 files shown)


toddr (approved):

cPholloway:
NOTE:  This failure actually shows that elevate succeeded:

https://github.com/cpanel/elevate/actions/runs/11298668305/job/31430465285?pr=524

So I am not going to bother rerunning the CI so that it can get a green check mark (there is a bug in the CI here)

davelcpanelnet:
@cPholloway please update this with elevate/main so we can pass the workflow checks.

PR #526, opened 2024-10-15
Support elevate for systems with Acronis backup agent

timmullin:
Case RE-575:

For Acronis, we must remove the Acronis backup agent package before the distro upgrade and re-install it after the distro upgrade.

Changelog: Support running elevate on systems that have the
  Acronis backup agent installed.

### DIFF
--- a/lib/Elevate/Components/Acronis.pm
+++ b/lib/Elevate/Components/Acronis.pm
@@ -1,0 +1,58 @@
+package Elevate::Components::Acronis;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::Acronis
+
+=head2 check
+
+noop
+
+=head2 pre_distro_upgrade
+
+Find out if Acronis is installed.
+If it is, uninstall it & make a note in the stage file.
+(We'll need to reinstall it after the OS upgrade.)
+
+=head2 post_distro_upgrade
+
+If the agent had been installed:
+Re-install the agent.
+
+=cut
+
+use cPstrict;
+
+use Elevate::Constants ();
+use Elevate::StageFile ();
+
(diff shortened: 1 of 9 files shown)


timmullin:
The installation instructions provided by the link referenced in the case says to download a script from them and run it.  I've inspected the contents of the script and it basically writes out and sets up the yum repo file and does "yum install -y acronis-backup-cpanel".  So, if they have the package installed, they would also have the repo.  That is why for the reinstall, I use dnf to reinstall the package instead of re-running the script.  I figure that is cleaner than downloading and running a script off of the internet that we do not control.

cPholloway (approved):
LGTM

lib/Elevate/Components/Acronis.pm
```diff
@@ -0,0 +1,58 @@
...
+Elevate::Components::Acronis
+
+=head2 check
+
+noop
+
+=head2 pre_distro_upgrade
+
+Find out:
+Is the Acronis agent installed?
```
davelcpanelnet, line 16: "Find out:" followed by a question seems rhetorical where we want technical.

"Installation Status of Acronis"

davelcpanelnet (changes requested):

lib/Elevate/Components/Acronis.pm
timmullin: I'll go ahead and make that change.

cPholloway (approved):

davelcpanelnet (approved):

timmullin:
I've added an update to squelch the warning about `acronis-backup-cpanel-1.8.3-818.el7.x86_64` being incompatible with AlmaLinux 8.

PR #529, opened 2024-10-19
Ensure the epel.repo file is "clean"

timmullin:
Case RE-812:

Customizations to /etc/yum.repos.d/epel.repo cause failures in the 4th stage of elevate. The customized repo file prevents AlmaLinux 8 version of the repo file from being installed.

Changelog: Fix elevate failures stemming from having a
  customized epel.repo file.

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -5123,6 +5123,7 @@ EOS
 
         $self->run_once("_disable_yum_plugin_fastestmirror");
         $self->run_once("_disable_known_yum_repositories");
+        $self->run_once("_fixup_epel_repo");
 
         return;
     }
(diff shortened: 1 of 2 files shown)


elevate-cpanel
```diff
@@ -5152,6 +5153,21 @@ EOS
         return;
     }
 
+    sub _fixup_epel_repo ($self) {
+
+        File::Copy::mv(
+            Elevate::Constants::YUM_REPOS_D . '/epel.repo',
```
toddr, line 5159: epel is not required on 7. We need to skip if it's not installed.

elevate-cpanel
```diff
@@ -5152,6 +5153,21 @@ EOS
...
+            Elevate::Constants::YUM_REPOS_D . '/epel.o.repo'
+        ) or die "Failed to rename epel.repo: $!";
+
+        $self->yum->reinstall('epel-release');
```
toddr, line 5163: is epel-release always a part of a repo?

toddr (changes requested):

elevate-cpanel
timmullin: I now test if `epel-release` is installed before proceeding.

elevate-cpanel
timmullin: First verifying that the package is installed should allow us to attempt a re-install.

toddr (approved):
I like this change. I will point out that it will lead to enabling epel where it wasn't before, but who cares? We're going to enable it for sure when they get to 8. I would just hilight we might want to test the scenario where it's installed but disabled on 7 and then we upgrade to 8. I can't imagine a way this would misbehave but.... ¯\_(ツ)_/¯

PR #530, opened 2024-10-23
Implement OpenStack Pruner

davelcpanelnet:
Implement OpenStack Pruner

Case RE-745: Implement OpenStack Pruner

Changelog:

### DIFF
--- a/.github/workflows/pruner.yml
+++ b/.github/workflows/pruner.yml
@@ -1,0 +1,12 @@
+name: Openstack Pruner
+run-name: Openstack Pruner
+on:
+  schedule:
+    - cron:  '5 * * * *'
+jobs:
+    run-pruner:
+      runs-on: [self-hosted]
+      steps:
+        - uses: actions/checkout@v4
+        - name: Run Openstack Pruner
+          run:  ./openstack/prune_openstack
--- a/.github/workflows/openstack/prune_openstack
+++ b/.github/workflows/openstack/prune_openstack
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+
+export OS_AUTH_TYPE=v3applicationcredential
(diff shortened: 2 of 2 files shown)


.github/workflows/pruner.yml
```diff
@@ -0,0 +1,12 @@
+name: Openstack Pruner
+run-name: Openstack Pruner
+on:
+  schedule:
+    - cron:  '*/5 */1 * * *'
```
toddr, line 5: we're going to run this every 5 minutes???

davelcpanelnet: “At every 5th minute past every hour.”

toddr: */5 means every 5 minutes
*/1 means every one hour

toddr (changes requested):
cron settings are not once a day they're currently every 5 minutes.

.github/workflows/pruner.yml
davelcpanelnet: Good geez this is why coding on 3 hours of sleep is a bad idea.  fixing

.github/workflows/pruner.yml
```diff
@@ -0,0 +1,12 @@
+name: Openstack Pruner
+run-name: Openstack Pruner
+on:
+  schedule:
+    - cron:  '0/5 */1 * * *'
```
toddr, line 5: */1 is redundant and confusing. Just do *

toddr: also not 0/5 just 5

`'5 * * * *'`

davelcpanelnet: OK Handled

cPholloway (approved):

toddr (approved):

PR #531, opened 2024-10-25
Add UnifiedLayer repos to vetted list

sloanebernstein:
Case RE-845: Add the `ul` repo and any repo that starts with `ul_` to the list of vetted repositories.

Changelog: Add 'ul' and 'ul_*' to the list of vetted repositories.


toddr (approved):

cPholloway (approved):
Code change looks good.  I am concerned about how leapp will handle this.  I would like to at least see a happy path go through with packages from these repos installed before we release this

PR #532, opened 2024-10-28
Assure DB integrity before elevation

timmullin:
Case RE-782:

Use mysqlcheck to check the integrity of their databases. Block if errors are found.

Changelog: Verify integrity of databases before elevation.

### DIFF
--- a/lib/Elevate/Components/MySQL.pm
+++ b/lib/Elevate/Components/MySQL.pm
@@ -36,6 +36,8 @@ use Cpanel::JSON                       ();
 use Cpanel::SafeRun::Simple            ();
 use Cpanel::DB::Map::Collection::Index ();
 use Cpanel::Exception                  ();
+use Cpanel::MysqlUtils::MyCnf::Basic   ();
+use Cpanel::MysqlUtils::Running        ();
 
 use Elevate::Database  ();
 use Elevate::Notify    ();
@@ -242,6 +244,7 @@ sub check ($self) {
     my $ok = 1;
     $ok = 0 unless $self->_blocker_old_mysql;
     $ok = 0 unless $self->_blocker_mysql_upgrade_in_progress;
+    $ok = 0 unless $self->_blocker_mysql_database_corrupted;
     $self->_warning_mysql_not_enabled();
     return $ok;
 }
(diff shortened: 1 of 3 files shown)


lib/Elevate/Components/MySQL.pm
```diff
@@ -360,6 +362,41 @@ sub _blocker_mysql_upgrade_in_progress ($self) {
...
+sub _blocker_mysql_database_corrupted ($self) {
+
+    # We cannot run mysqlcheck if mysql is not running
+    if ( !Cpanel::MysqlUtils::Running::is_mysql_running() ) {
+
+        if ( Cpanel::Services::Enabled::is_enabled('mysql') ) {
+            my $db_version = Elevate::Database::get_local_database_version();
+            my $db_type    = Elevate::Database::get_database_type_name_from_version($db_version);
+
+            WARN("$db_type is not running; this prevents us from verifying the integrity of the databases.");
```
cPholloway, line 374: FIX: Instead of throwing a warning and returning, we should attempt to restart the database server here and block if it fails to start.

lib/Elevate/Components/MySQL.pm
```diff
@@ -360,6 +362,41 @@ sub _blocker_mysql_upgrade_in_progress ($self) {
...
+    }
+
+    # Perform a medium check on all databases and only output errors
+    my $mysqlcheck_path = Cpanel::Binaries::path('mysqlcheck');
+    my $output          = $self->ssystem_capture_output( $mysqlcheck_path, '-c', '-m', '-A', '--silent' );
+
+    # mysqlcheck doesn't return an error code
+    # We check for lines that actually begin with "Error" (or "error")
+    # because we don't want to block of only warnings are found
+    if ( scalar grep { /^[eE]rror/ } @{ $output->{stdout} } ) {
```
cPholloway, line 387: CONSIDER: Ignore case in the regex here so that we catch ERROR here too.  Not sure if this is necessary.

cPholloway (changes requested):

lib/Elevate/Components/MySQL.pm
timmullin: Okay, can do.

lib/Elevate/Components/MySQL.pm
timmullin: Yeah, I think that is a better approach.

lib/Elevate/Components/MySQL.pm
```diff
@@ -360,6 +363,51 @@ sub _blocker_mysql_upgrade_in_progress ($self) {
...
+
+        my $output = $self->ssystem_capture_output(qw{/scripts/restartsrv_mysql});
+
+        WARN('Database server was down, starting it to check database integrity');
+        unless ( Cpanel::MysqlUtils::Running::is_mysql_running()
+            && grep { /mysql (?:re)?started successfully/ } @{ $output->{stdout} } ) {
+
+            return $self->has_blocker( <<~"EOS" );
+            We have not been able to successfully start the database server
+            to check the integrity of the databases.
```
cPholloway, line 386: FIX: Lets improve this message to be more helpful to the customer.  Maybe something like:

```
Unable to to start the database server to check database integrity.  Additional information can be found in the error log located at ...

To attempt to start the database server, execute ...
```

cPholloway (changes requested):

cPholloway (approved):
Approved pending CI passing

toddr (approved):

PR #539, opened 2024-10-30
Re-install Softaculous after upgrade

sloanebernstein:
Case RE-846: It was reported that the Softaculous plugin for cPanel was displaying blank pages after ELevate, and that a re-install of the plugin fixed this issue. That seems reasonable, so detect Softaculous and re-install if present.

Changelog: Re-install Softaculous after upgrade.

### DIFF
--- a/lib/Elevate/Components/Softaculous.pm
+++ b/lib/Elevate/Components/Softaculous.pm
@@ -1,0 +1,84 @@
+package Elevate::Components::Softaculous;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::Softaculous
+
+=head2 pre_distro_upgrade
+
+If we can query the version of Softaculous through their CLI tool, it is installed.
+
+=head2 post_distro_upgrade
+
+If Softaculous is installed, re-install it after the upgrade.
+
+NOTE: This needs to happen after cPanel is updated to work with the new OS, since Softaculous relies on cPanel PHP.
+
+=cut
+
+use cPstrict;
+
+use Elevate::Fetch     ();
+use Elevate::StageFile ();
+
+use Cpanel::Binaries        ();
+use Cpanel::SafeRun::Object ();
+
+use Log::Log4perl qw(:easy);
+
+use parent qw{Elevate::Components::Base};
+
+use Simple::Accessor qw(cli_path);
+
+sub _build_cli_path { return '/usr/local/cpanel/whostmgr/docroot/cgi/softaculous/cli.php' }
+
+sub pre_distro_upgrade ($self) {
+
+    return unless -r $self->cli_path;
+
+    my $sr = _run_script( $self->cli_path );
+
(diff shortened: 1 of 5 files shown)


lib/Elevate/Components/Softaculous.pm
```diff
@@ -0,0 +1,75 @@
...
+        program => Cpanel::Binaries::path('php'),
+        args    => [ $SOFTACULOUS_CLI_PATH, '--version' ],
+    );
+
+    return if $sr->exec_failed() || $sr->to_exception();
+
+    my $version = $sr->stdout();
+    chomp $version;
+
+    if ($version) {
```
toddr, line 49: length $version

lib/Elevate/Components/Softaculous.pm
```diff
@@ -0,0 +1,75 @@
...
+        INFO('Softaculous has been detected. The system will re-install that software after the distro upgrade.');
+        Elevate::StageFile::update_stage_file( { softaculous => $version } );
+    }
+
+    return;
+}
+
+sub post_distro_upgrade ($self) {
+
+    my $version = Elevate::StageFile::read_stage_file( 'softaculous', '' );
```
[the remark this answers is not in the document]
cPholloway: The second argument is the default value to return if the 1st argument is missing from the stagefile

lib/Elevate/Components/Softaculous.pm
```diff
@@ -0,0 +1,75 @@
...
+    return unless $version;
+
+    my $path = Elevate::Fetch::script( 'https://files.softaculous.com/install.sh', 'softaculous_install' );
+
+    if ($path) {
+        INFO('Re-installing Softaculous:');
+        LOGDIE('Re-installation of Softaculous failed.') if !$self->ssystem( $path, '--reinstall' );
```
cPholloway, line 66: I don't think that this failing should be fatal for the script.  Can we provide the customer with some information on how to reinstall softaculous to get it working again and add this to the final notifications at the end instead?

cPholloway (changes requested):
Please add a unit test for the new component

lib/Elevate/Components/Softaculous.pm
sloanebernstein: I went back and forth about whether it should be fatal or not. Changing it back.

toddr: Is it a problem they can clean up after Elevate completes? Continue. If it's going to break the rest of elevate then stop and make them fix it.

cPholloway: > Is it a problem they can clean up after Elevate completes?

Yes, this is what customers are having to do now.  The released version of Elevate completes successfully now without any of this in place.

cPholloway (dismissed):
Approved pending follow up case to finish the unit test. Needs assertions for `post_distro_upgrade`

cPholloway (approved):

PR #542, opened 2024-10-31
Ensure final dnf update is executed

cPholloway:
Case RE-934: In RE-252, a bug was introduced that resulted in the final dnf update spitting out help text instead of updating the system.  This change updates the command to be valid again so that the final update takes place.

Changelog: Ensure final dnf update is executed

### DIFF
--- a/lib/Elevate/PkgMgr/YUM.pm
+++ b/lib/Elevate/PkgMgr/YUM.pm
@@ -254,13 +254,16 @@ sub update_with_options ( $self, $options, $pkgs ) {
     return;
 }
 
-sub update_allow_erasing ( $self, @args ) {
+sub update_allow_erasing ( $self, @additional_args ) {
     my $pkgmgr = $self->_pkgmgr();
 
-    my @additional_args = scalar @args ? @args : '';
+    my @args = (
+        '-y',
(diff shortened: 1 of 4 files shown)


toddr (approved):

lib/Elevate/PkgMgr/YUM.pm
```diff
@@ -254,13 +254,16 @@ sub update_with_options ( $self, $options, $pkgs ) {
...
-    my @additional_args = scalar @args ? @args : '';
-
-    $self->ssystem( $pkgmgr, '-y', '--allowerasing', @additional_args, 'update' );
+    my @args = (
+        '-y',
+        '--allowerasing',
+    );
 
+    push @args, @additional_args;
+    $self->ssystem_and_die( $pkgmgr, @args, 'update' );
```
toddr, line 266: so the only change is `ssystem_and_die` ?

cPholloway: There was the potential for `@args` to be `''` when no args were passed into `update_allow_erasing`.  For whatever reason, the empty string was causing the call to dnf to fail so I just made it to where that empty string could never be present.

PR #550, opened 2024-11-14
Add 'sys-snap' to the list of packages that PackageRestore handles

cPholloway:
Case RE-952: This adds 'sys-snap' to the list of packages that the PackageRestore component handles.  It also now explicitely removes the packages as it was previously relying on leapp to implicitely remove them.

Changelog: Add 'sys-snap' to the list of packages that the
 PackageRestore component handles

### DIFF
--- a/elevate-cpanel
+++ b/elevate-cpanel
@@ -4354,6 +4354,7 @@ EOS
     sub _get_packages_to_check () {
         return qw{
           net-snmp
+          sys-snap
         };
     }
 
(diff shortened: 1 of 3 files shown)


elevate-cpanel
```diff
@@ -4368,6 +4369,8 @@ EOS
             }
         }
 
+        Elevate::PkgMgr::remove(@installed_packages);
```
toddr, line 4372: is it a bug this wasn't here before?

toddr (approved):

cPholloway:
I'm going to update a unit test for this.  Noticed that something needs a mock when I was checking something this morning

davelcpanelnet (approved):

PR #552, opened 2024-11-15
Beef up integration test

troglodyne:
case RE-948: I did the best here to be terse but understandable, which was difficult without the QA tools. Regardless, I think this will work.

### DIFF
--- a/t/integration/complete.t
+++ b/t/integration/complete.t
@@ -8,16 +8,73 @@
 use lib '/usr/local/cpanel/';
 
 use Cpanel::OS              ();
+use Cpanel::JSON            ();
 use Cpanel::SafeRun::Simple ();
 
-use Test::More tests => 7;
+use Test::More;
 
-use Data::Dumper;
+# "Poor man's" FailWarnings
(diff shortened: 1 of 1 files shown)


troglodyne:
Presumably if the pipeline passes (or fails legitimately) I'll have done my job

cPholloway:
You'll need to add `Test::FailWarnings` to `t/cpanfile` so that it gets installed in the setup step.  Can't believe we don't already use that :\

toddr:
> You'll need to add `Test::FailWarnings` to `t/cpanfile` so that it gets installed in the setup step. Can't believe we don't already use that :\

Or make this Test2
```perl
use cPstrict;
use FindBin;

use Test2::V0;
use Test2::Tools::Explain;
use Test2::Plugin::NoWarnings;
use Test2::Tools::Exception;

use Test::MockModule qw/strict/;

use lib $FindBin::Bin . "/lib";
```

cPholloway (approved):

FRooter (approved):

PR #553, opened 2024-11-15
Avoid Perl error in 'Elevate::Components::Repositories'

cPholloway:
Case RE-964: This change adds a better guard in
'Elevate::Components::Repositories::_autofix_duplicate_repoids'. It also initializes '$self->{__duplicate_repoids} as a href instead of an aref.

Changelog: Avoid Perl error in 'Elevate::Components::Repositories'


toddr (approved):

sloanebernstein (approved):
Tests missed this because they don't cover `check` or `_blocker_invalid_yum_repos`.

davelcpanelnet (approved):

PR #556, opened 2024-11-19
Update DiskSpace check to account for Ubuntu

cPholloway:
Case RE-672: This increasing the amount of free space needed on '/boot'. It also adds a check for '/tmp' and '/' ensuring that they have a minimally amount of free space as well.

Changelog: Update DiskSpace check to account for Ubuntu


toddr:
How does this allow for securetmp?

cPholloway:
> How does this allow for securetmp?

This should work fine with securetmp.  It is just checking the output of `/usr/bin/df -hk / /boot /tmp /usr/local/cpanel /var/lib`

PR #559, opened 2024-11-21
Prefer using apt-get over apt in Elevate::PkgMgr

cPholloway:
Case RE-980: apt will throw a warning to STDERR that it is not intended for scripts which causes unnecessary noise in the elevate logs.  This change updates Elevate::PkgMgr to prefer apt-get over apt in order to avoid this warning and help make parsing output more consistent

Changelog:


troglodyne:
I made this same mistake when doing initial Ubuntu support. You may want to look up the ENV vars you need too from `Cpanel/PackMan/Sys/apt.pm`.

toddr (approved):

PR #567, opened 2024-12-03
Add Ubuntu support for bootloader test

sloanebernstein:

### DIFF
--- a/lib/Elevate/Components/Grub2ControlTest.pm
+++ b/lib/Elevate/Components/Grub2ControlTest.pm
@@ -1,0 +1,147 @@
+package Elevate::Components::Grub2ControlTest;
+
+=encoding utf-8
+
+=head1 NAME
+
+Elevate::Components::Grub2ControlTest
+
+=head2 mark_cmdline
+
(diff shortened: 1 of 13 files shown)


lib/Elevate/Components/Grub2ControlTest.pm
```diff
@@ -0,0 +1,200 @@
...
+    return $id if $id;
+
+    $id = int( rand(100000) );
+    Elevate::StageFile::update_stage_file( { 'bootloader_random_tag', $id } );
+    return $id;
+}
+
+sub _add_kernel_arg ( $self, $arg ) {
+
+    if ( Elevate::OS::bootloader_config_method() eq 'grubby' ) {
```
sloanebernstein, line 78: It can be argued that this is a violation of the `Cpanel::OS` principles which are (probably) carried over into `Elevate::OS`. The counterargument is that the correct way to handle this, which is to create a submodule for each set of behaviors, sounds like a lot of boilerplate with not so much meat.

cPholloway (approved):

PR #571, opened 2024-12-05
Add component to manage port 1022 for do-release-upgrade

cPholloway:
Case RE-943: When performing an elevation from u20->u22, we rely on the do-release-upgrade script to perform the OS upgrade.  do-release-upgrade will create a secondary SSH connection on port 1022 as a fallback mechanism in the event that the primary SSH connection is disrupted during the upgrade process. This change ensures that port 1022 is open before executing do-release-upgrade so that the fallback method will work if it is required during the upgrade.  It also ensures that port 1022 is closed once the upgrade is complete if it was closed before the upgrade started.

Changelog: Add component to manage port 1022 for ELevations that rely on
 the do-release-upgrade script to perform the OS upgrade


troglodyne:
While I'm sure someone out there rips out UFW on principle, the grand majority will not. Looks like it ought to work.

troglodyne (dismissed):

toddr (approved):

PR #583, opened 2024-12-18
RE-1001: Move GitHub Secrets into Variables that have no Secrecy need

davelcpanelnet:
Case RE-1001: Move GitHub Secrets into Variables that can be

Changelog:


davelcpanelnet:
The PR will serve as a walk through for what will be needed in terms of configuration of Secrets and Variables when we move to Enterprise.  Those will need to be in place here before the proper OpenStack Spawns will take place.

PR #587, opened 2025-02-04
Do not restore config files for dropped packages

cPholloway:
Case RE-1179: Previously, we did not filter out dropped packages from the list of EA4 packages that had config files that needed restored. This could result in odd bugs where a file was put in place that should not have been.

Changelog: Do not restore config files for dropped packages


troglodyne (approved):

toddr (approved):

toddr:
LGTM

PR #596, opened 2025-03-17
Update CloudLinux Elevate repository names

prilr:
CloudLinux package repo names in the cloudlinux-elevate-release package have changed, but the names elevate-cpanel is using were not updated.

Normally it’s not a problem during standard conversion, but when trying to use the leappbeta flag to enable CL's beta repo, it doesn’t work correctly anymore.

This PR updates the names.


toddr:
we've moved to an internal repo. we'll get back to you.

